PR #9, opened 2024-04-21
add admin and baseline network policy

aojea:
It looks relatively simple to implement


pkg/networkpolicy/adminnetworkpolicy.go
```diff
@@ -0,0 +1,239 @@
...
+		msg += fmt.Sprintf(" SrcPod (%s/%s): %d AdminNetworkPolicy", srcPod.Name, srcPod.Namespace, len(srcPodAdminNetworkPolices))
+	}
+	if dstPod != nil {
+		msg += fmt.Sprintf(" DstPod (%s/%s): %d AdminNetworkPolicy", dstPod.Name, dstPod.Namespace, len(dstPodAdminNetworkPolices))
+	}
+	klog.V(2).Infof("%s", msg)
+
+	// For a connection from a source pod to a destination pod to be allowed,
+	// both the egress policy on the source pod and the ingress policy on the
+	// destination pod need to allow the connection.
```
npinaeva, line 37: I think it is a bit more complicated than that. A condition for a successful connection (ignoring baseline ANP) is something like
`egress Allow && ingress Allow` =
`(egress ANP Allow || egress ANP skip && egress NP Allow) && (ingress ANP Allow || ingress ANP Skip && ingress NP Allow)`
An example match may be:
egress ANP Skip, egress NP Allow => egress allow
ingress ANP Allow, ingress NP Drop => ingress allow
With current logic it will be ANP verdict => Skip, NP verdict =>Drop, result => Drop.

So with admin network policies we may need to get verdicts for ingress and egress separately and then evaluate an expression like this ^

aojea: I see this simpler as 2 checks, 

check 1: ANP it can be "not apply", allowed, skipped or denied

check 2. Network policies, only comes here if 1 does not apply or is skipped, no?

pkg/networkpolicy/adminnetworkpolicy.go
```diff
@@ -0,0 +1,239 @@
...
+						podSelector(&from.Pods.PodSelector, srcPod) {
+						return rule.Action
+					}
+				}
+			}
+
+		}
+	}
+
+	return npav1alpha1.AdminNetworkPolicyRuleActionPass
```
tssurya, line 129: ah so no ANPs match for the pod is considered as a pass so that netpol's can be evaluated?

aojea: exactly https://github.com/kubernetes-sigs/kube-network-policies/pull/9#discussion_r1575377948

pkg/networkpolicy/adminnetworkpolicy.go
npinaeva: I think some unit tests for anp will help us understand this case better

aojea: yep, yep

aojea: almost there, half way with the tests

pkg/networkpolicy/adminnetworkpolicy.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	// Traffic is allowed in both direction
+	if ingressAction == npav1alpha1.AdminNetworkPolicyRuleActionAllow &&
+		egressAction == npav1alpha1.AdminNetworkPolicyRuleActionAllow {
+		return npav1alpha1.AdminNetworkPolicyRuleActionAllow
+	}
+
+	// Traffic has to pass to be passed to Network Policies
+	return npav1alpha1.AdminNetworkPolicyRuleActionPass
```
aojea, line 58: @npinaeva @tssurya see the logic here, 
1. we evaluate if the srcPod has any policy in egress to the dstPod and get its action
2. we evaluate if the dstPod has any policy in ingress from the srcPod and get its action
3. If there is at least one deny we just drop
4. if both are allow then we allow
5. the rest means it has to passed down to keep evaluating by network policies

pkg/networkpolicy/adminnetworkpolicy.go
```diff
@@ -0,0 +1,282 @@
...
+		if pod.Spec.NodeName == node.Name {
+			return true
+		}
+	}
+	return false
+}
+
+// getAdminNetworkPoliciesForPod returns the list of Admin Network Policies matching the Pod
+// The list is ordered by priority, from higher to lower.
+func (c *Controller) getAdminNetworkPoliciesForPod(pod *v1.Pod) []*npav1alpha1.AdminNetworkPolicy {
```
npinaeva, line 206: I find it slightly confusing, that `getNetworkPoliciesForPod` only returns policies from the same namespace, and then it checks if they really apply to the pod in `evaluator`, but here we check it all at once. 
Maybe we could change `getNetworkPoliciesForPod` to also check podSelector?

pkg/networkpolicy/adminnetworkpolicy.go
npinaeva: I think we are still missing some allow conditions, mainly when one direction ANP says Pass and another says Allow. To check that we need a test where both ANPs and NPs are present, example from the previous comment should still work.

pkg/networkpolicy/adminnetworkpolicy.go
aojea: this is evaluating the Spec.Subject specifically , can you expand what else need to be evaluated?

pkg/networkpolicy/controller.go
```diff
@@ -298,6 +326,75 @@ func (c *Controller) Run(ctx context.Context) error {
 	return nil
 }
 
+// evaluatePacket evalute the network policies using the following order:
+// 1. AdminNetworkPolicies in Egress for the source Pod/IP
+// 2. NetworkPolicies in Egress (if needed) for the source Pod/IP
+// 3. AdminNetworkPolicies in Ingress for the destination Pod/IP
+// 4. NetworkPolicies in Ingress (if needed) for the destination Pod/IP
+func (c *Controller) evaluatePacket(p packet) bool {
```
aojea, line 334: @npinaeva please if this logic is correct now

aojea: ref https://github.com/kubernetes-sigs/kube-network-policies/pull/9#discussion_r1577518558

npinaeva: yeah I think looks correct now!

pkg/networkpolicy/adminnetworkpolicy.go
npinaeva: I am comparing similar methods for networkpolicy and adminnetworkpolicy.
This one for adminnetworkolicy is actually good, but networkpolicy `getNetworkPoliciesForPod` method doesn't filter based on podSelector. So can we add that filtering here https://github.com/kubernetes-sigs/kube-network-policies/blob/main/pkg/networkpolicy/networkpolicy.go#L19-L23 and remove from here https://github.com/kubernetes-sigs/kube-network-policies/blob/main/pkg/networkpolicy/networkpolicy.go#L78-L87?

aojea: great review

aojea: @npinaeva do you mind sending a PR for that?

Issue #12, opened 2024-04-22
[Flaky] Services should create endpoints for unready pods - Netfilter Bug


aojea:
It flakes on the network policy jobs https://testgrid.k8s.io/sig-network-gce#network-policies,%20google-gce

Analyzing one occurrence https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-e2e-gce-network-policies/1782274345247182848

I can see DNS request on the trace but never a try to connect to the Service, that may indicate that something is dealing with the DNS request or response

aojea:
I can reproduce the problem, it seems that when using curl sometimes it take too long to answer and times out ... surprisingly if I use ping or nslookup it replies fast ...

aojea:
```
time nslookup tolerate-unready.services-5834.svc.cluster.local
Server:         10.0.0.10
Address:        10.0.0.10#53

Name:   tolerate-unready.services-5834.svc.cluster.local
Address: 10.0.35.116


real    0m0.389s
user    0m0.009s
sys     0m0.011s
bash-5.0# curl -w "time_namelookup: %{time_namelookup}\n" http://tolerate-unready.services-5834.svc.cluster.local:80/
curl: (7) Failed to connect to tolerate-unready.services-5834.svc.cluster.local port 80 after 5060 ms: Connection refused
time_namelookup: 5.059978
bash-5.0# curl -w "time_namelookup: %{time_namelookup}\n" http://tolerate-unready.services-5834.svc.cluster.local:80/
curl: (7) Failed to connect to tolerate-unready.services-5834.svc.cluster.local port 80 after 15025 ms: Connection refused
time_namelookup: 15.025116
```

it is only with curl, nslookup works perfectly and fast

```
 curl --version
curl 7.79.1 (x86_64-alpine-linux-musl) libcurl/7.79.1 OpenSSL/1.1.1n zlib/1.2.12 nghttp2/1.41.0
Release-Date: 2021-09-22
Protocols: dict file ftp ftps gopher gophers http https imap imaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS HSTS HTTP2 HTTPS-proxy IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets
```

```
cat /etc/resolv.conf
[...]

PR #13, opened 2024-04-23
drop iptables support

aojea:
no time for dealing with the old stuff , since we are moving on with nftables everywhere , having to maintain both is a lot of work for no much return

https://github.com/kubernetes-sigs/kube-network-policies/issues/12

### DIFF
--- a/pkg/networkpolicy/controller.go
+++ b/pkg/networkpolicy/controller.go
@@ -47,7 +47,7 @@ import (
 // https://netfilter.org/projects/libnetfilter_queue/doxygen/html/
 
 const (
-	controllerName = "kube-netpol"
+	controllerName = "kube-network-policies"
 	podIPIndex     = "podIPKeyIndex"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/networkpolicy/controller.go
```diff
@@ -56,28 +51,9 @@ type Config struct {
...
-	}
-	if strings.Contains(out.String(), "legacy") {
-		return true
-	}
-	return false
-}
-
 // NewController returns a new *Controller.
 func NewController(client clientset.Interface,
+	nft knftables.Interface,
```
danwinship, line 56: Why pass `nft` into `NewController` but not `nfq`?

I feel like it makes more sense to have `Controller` create the knftables object itself. It knows what it needs; `main` doesn't.

aojea: to mock it in the tests

aojea: also nfq needs to register the hooks that depend on the controller

aojea:
this is working on ci https://github.com/kubernetes/kubernetes/pull/124484

PR #14, opened 2024-04-24
networkpolicy: Move subject filtering logic to getNetworkPoliciesForPod.

npinaeva:
I removed this log https://github.com/kubernetes-sigs/kube-network-policies/blob/main/pkg/networkpolicy/networkpolicy.go#L85 for symmetry with adminnetworkpolicy code, lmk if I should add it back.


aojea:
/lgtm
/approve

Thanks

PR #17, opened 2024-04-26
Define Goals and non Goals of the project

aojea:
clarify the project scope

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ Covered: New features implemented under feature gates for fast iteraton
 
 Implementing functionality that is not part of an official or proposed Kubernetes network policy API
 
+Contributions aligned with the project scope, feedback and bugs reports are very welcome, but this is an Open Source project maintained by the community with a best-effort support model.
 ## References
 
 * https://home.regit.org/netfilter-en/using-nfqueue-and-libnetfilter_queue/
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +63,29 @@ See [TESTING](docs/testing/README.md)
...
+Covered Work: e2e testing, scalability, performance and reliability
+
+P0: Provide Early Feedback
+
+Stakeholders: SIG Network, Network Policy API Working Group
+Covered: New features implemented under feature gates for fast iteraton
+
+### Non Goals
+
+Implementing additional functionalities unrelated to Network Policies
```
danwinship, line 87: "Implementing functionality that is not part of an official or proposed Kubernetes network policy API" ?

README.md
```diff
@@ -63,6 +63,29 @@ See [TESTING](docs/testing/README.md)
...
+
```
danwinship, line 88: "Supporting production use" should be either a Goal or a Non-Goal. (ie, how do we handle bug reports?)

Likewise, "supporting use at scale"

aojea: I'm trying to indicate that people is free to use it as they want and feedback is really appreciated:  "given enough eyeballs, all bugs are shallow", but this is an open source project with best effort support ...  the main goal of the project is zero bugs ... best performance ... same as we do with KIND ... don't know how to write that

README.md
aojea, line 87:
```suggestion
Implementing functionality that is not part of an official or proposed Kubernetes network policy API
```

README.md
aojea: will clarify in the scope

Issue #10, opened 2024-04-21
Do not process packet twice


aojea:
1. PodA and PodB in same node is only evaluated once, since it is in forwarding and return will not hit the new Contrack state (I think)
2. PodA in NodeA to PodB in NodeB is evaluated in both Nodes
3. PodA in NodeA to external , is only evaluated once
4. External to PodA in nodeA is only evaluated once

If a network policy was evaluated on origin, don't evaluate again in the destination.
 
/kind feature

Rahul-D78:
Hey aojea, I'm interested to work on this. Any pointers on where to start in the codebase would be greatly appreciated. Thanks!

Issue #31, opened 2024-05-24
Only NFQUEUE a nftables set of IPs matching the selectors?


vaskozl:
> There are also some performance improvements that can be applied, such as to restrict the packets that are sent to userspace to the ones that have network policies only. This effectively means that network policies are applied ONLY at the time the connection is initatied by whatever the conntrack kernel understand by NEW connection.

I found this comment quite interesting as I was interested in an `nftables` network policies implementation. We could this with by creating a a set:

```
for _, ndp := range podSelectorEndpoints {
        tx.Add(&knftables.Element{
                Set: netpoledPodsSetName,
                Key: []string{string(ndp.PodIP)},
        })
}
```

Then we can use two queues, one for when the PodIP is the source and one when it's the destination, so we only need to evaluate egress and ingress respectively.

Something like:
```
tx.Add(&knftables.Rule{
        Chain: filterInputChain,
        Rule: knftables.Concat(
                ipX, "saddr", "@", netpoledPodsSetName, "queue", "num", fmt.Sprintf(c.config.IngressQueueID),
        ),
        Comment: ptr.To("Evaluate ingress network policies"),
})
```

This partially solve the double processing in https://github.com/kubernetes-sigs/kube-network-policies/issues/10, assuming the `@netpoledPodsSet` set only includes pods currently running on the node. (We'd still process in userspace twice but we'd do only half of the checks).

While this seems good in theory to me, I can't help but wonder whether the added complexity here is worth it, because now we have to maintain an up to date set of on each node.

We could actually maintain a set per network policy (instead of one big one). If we do all that, then it's not much of a strech to also maintain an ingress and egress nftables set per netpol each contain addr/protocol and port(range)?

Something like:

```
tx.Add(&knftables.Rule{
        Chain: ingressEnabledChain,
        Rule: knftables.Concat(
                ipX, "saddr", ".", "dport", "@", ingressSetName, "ip", "daddr", "@", netpoledPodsSetName, "accept"
        ),
        Comment: ptr.To("Allow defined ingress"),
})
```

(We'd actually also need a `noPort` ingress/egress set and rule, as inet_service must be a port or range)

TLDR: If we bother to maintain one set, should we not maintain 5 and do everything in `nftables` (future features aside)?

aojea:
those were my thoughts when I added the comment, and I kind of like the simplicity of existing solution that just needs one nftables rule ... I think that is a matter of taking numbers at this point

[...]

PR #39, opened 2024-06-23
Only process traffic impacted by network policies

aojea:
optimize the datapath not having to send all packets to user space, only the ones that are subject of network policies

Fixes: https://github.com/kubernetes-sigs/kube-network-policies/issues/10, #31, #12

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,11 +64,17 @@ func main() {
 		klog.Fatalf("error parsing metrics bind address %s : %v", metricsBindAddress, err)
 	}
 
+	nodeName := os.Getenv("MY_NODE_NAME")
+	if nodeName == "" {
+		klog.Fatalf("node name not set, please set the environment variable using the Downward API")
+	}
+
 	cfg := networkpolicy.Config{
 		AdminNetworkPolicy:         adminNetworkPolicy,
 		BaselineAdminNetworkPolicy: baselineAdminNetworkPolicy,
 		FailOpen:                   failOpen,
 		QueueID:                    queueID,
+		NodeName:                   nodeName,
 	}
 	// creates the in-cluster config
 	config, err := rest.InClusterConfig()
--- a/pkg/networkpolicy/networkpolicy.go
+++ b/pkg/networkpolicy/networkpolicy.go
@@ -10,6 +10,29 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// getPodsForNetworkPolicyNode obtains all the Pods selected by the network policy for current Node
+func (c *Controller) getLocalPodsForNetworkPolicy(networkPolicy *networkingv1.NetworkPolicy) []*v1.Pod {
+	if networkPolicy == nil {
+		return nil
+	}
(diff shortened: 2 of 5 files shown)


aojea:
/assign @danwinship 

Dan please take a look

pkg/networkpolicy/networkpolicy.go
```diff
@@ -10,6 +10,29 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// getPodsForNetworkPolicyNode obtains all the Pods selected by the network policy for current Node
+func (c *Controller) getLocalPodsForNetworkPolicy(networkPolicy *networkingv1.NetworkPolicy) []*v1.Pod {
+	if networkPolicy == nil {
+		return nil
+	}
+	podSelector, err := metav1.LabelSelectorAsSelector(&networkPolicy.Spec.PodSelector)
```
danwinship, line 18: So FTR, with the existing code, if you have Pod A on Node 1 that denies all ingress, and Pod B on Node 2 tries to send to it, then Node 2 will drop the packets. With this new code, Node 2 would instead forward the packet to Node 1, and then Node 1 would drop it.

cmd/main.go
```diff
@@ -64,11 +64,17 @@ func main() {
 		klog.Fatalf("error parsing metrics bind address %s : %v", metricsBindAddress, err)
 	}
 
+	nodeName := os.Getenv("MY_NODE_NAME")
+	if nodeName == "" {
+		klog.Fatalf("node name not set, please set the environment variable using the Downward API")
```
danwinship, line 69: It would be better to make this a command-line option, and for it to be optional.

danwinship (commented):
This looks like it will work (though you lose all the optimization once you add even a single ANP)...

Another possibility would be to just add "local detector" options like kube-proxy (`--cluster-cidr`, use node `podCIDRs`, etc)

pkg/networkpolicy/networkpolicy.go
aojea: yeah, is the trade off, but at least the rules are only evaluated once in one node, previously allowed rules were evaluated twice

cmd/main.go
aojea: but now the reconciliation logic depends on the nodename and the installation yaml defines the downward API variable ... is there any case you see people may want to run without this ?

aojea:
> This looks like it will work (though you lose all the optimization once you add even a single ANP)...

Cluster Wide Policies are hard to implement in the dataplane

> Another possibility would be to just add "local detector" options like kube-proxy (`--cluster-cidr`, use node `podCIDRs`, etc)

Since the controller already has the pod information this looks simpler, kube-proxy does not watch pods and need to depend on those heuristics

danwinship:
It would still be better for it to be a command-line option, even if it's required. Required environment variables are terrible.

aojea:
> It would still be better for it to be a command-line option, even if it's required. Required environment variables are terrible.

I just want to fix the flake, let me open an issue to not forget and do the change

Issue #43, opened 2024-06-25
Use a command line option to pass the node name


aojea:
The node name is passed as an environment variable
           

It would still be better for it to be a command-line option, even if it's required. Required environment variables are terrible.

_Originally posted by @danwinship in https://github.com/kubernetes-sigs/kube-network-policies/issues/39#issuecomment-2189358129_

PR #44, opened 2024-06-26
don't use obscure env variables and use flags

aojea:
Let's do the things the right way

Fixes: #43


cmd/main.go
```diff
@@ -64,9 +67,9 @@ func main() {
 		klog.Fatalf("error parsing metrics bind address %s : %v", metricsBindAddress, err)
 	}
 
-	nodeName := os.Getenv("MY_NODE_NAME")
-	if nodeName == "" {
-		klog.Fatalf("node name not set, please set the environment variable using the Downward API")
+	nodeName, err := nodeutil.GetHostname(hostnameOverride)
```
aojea, line 70: this is exactly how kube-proxy does

install-anp.yaml
```diff
@@ -80,7 +80,7 @@ spec:
       - name: kube-network-policies
         image: registry.k8s.io/networking/kube-network-policies:v0.3.0
         command: ["/bin/netpol"]
-        args: ["-v=2", "-admin-network-policy=true", "-baseline-admin-network-policy=true"]
+        args: ["-v=2", "-admin-network-policy=true", "-baseline-admin-network-policy=true","--hostname-override=${MY_NODE_NAME}"]
```
aojea, line 83: the main problem because I added the env variables is because I was not sure you can do this, of using the env variable as the flag value and I was too lazy ... now that you mentioned I can see this is exactly how most people do https://grep.app/search?q=hostname-override%3D%24&filter[lang][0]=YAML

cmd/main.go
```diff
@@ -64,9 +67,9 @@ func main() {
...
+	if err != nil {
+		klog.Fatalf("can not obtain the node name, use the hostname-override flag if you want to set it to a specific value")
```
danwinship, line 72: include `err`?

danwinship:
/lgtm
/hold
if you want to update the error message but maybe that was intentional

cmd/main.go
aojea: done

install-anp.yaml
aojea: ok, embarrassing things https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

PR #47, opened 2024-07-01
Add an output-chain, needed for proxy-mode=ipvs

uablrek:
If the receiving pod is local, then ipvs sends packet via the output hook (not forward). So both are needed

Fixes #46 

Example: 
```
# nft list table inet kube-network-policies
table inet kube-network-policies {
        comment "rules for kubernetes NetworkPolicy"
        set podips-v4 {
                type ipv4_addr
                comment "Local V4 Pod IPs with Network Policies"
                elements = { 11.0.2.2, 11.0.2.3,
                             11.0.2.4 }
        }

        set podips-v6 {
                type ipv6_addr
                comment "Local V6 Pod IPs with Network Policies"
                elements = { 1100::202,
                             1100::203,
                             1100::204 }
        }

        chain forward {
                type filter hook forward priority filter - 5; policy accept;
                ct state established,related accept
                ip saddr @podips-v4 queue to 100 comment "process IPv4 traffic with network policy enforcement"
                ip daddr @podips-v4 queue to 100 comment "process IPv4 traffic with network policy enforcement"
                ip6 saddr @podips-v6 queue to 100 comment "process IPv6 traffic with network policy enforcement"
                ip6 daddr @podips-v6 queue to 100 comment "process IPv6 traffic with network policy enforcement"
        }

        chain output {
                type filter hook output priority filter - 5; policy accept;
                ct state established,related accept
                ip saddr @podips-v4 queue to 100 comment "process IPv4 traffic with network policy enforcement"
                ip daddr @podips-v4 queue to 100 comment "process IPv4 traffic with network policy enforcement"
                ip6 saddr @podips-v6 queue to 100 comment "process IPv6 traffic with network policy enforcement"
                ip6 daddr @podips-v6 queue to 100 comment "process IPv6 traffic with network policy enforcement"
        }
}
```
The "chain output" is added by this PR.

Please see issue #46 for more info

Issue #15, opened 2024-04-25
Header processing is too simplified for IPv6


uablrek:
The code for extracting ports, and a pointer to the payload is too simplified:
https://github.com/kubernetes-sigs/kube-network-policies/blob/c7fc563db4e9f011d200da021b59098ac02e9bdb/pkg/networkpolicy/packet.go#L77-L84

It assumes that there are no extension headers. This is mentioned in:
https://github.com/kubernetes-sigs/kube-network-policies/blob/c7fc563db4e9f011d200da021b59098ac02e9bdb/pkg/networkpolicy/packet.go#L57

But should be documented, as it is a serious limitation. It must eventually be addressed, and then the copied packet length in:
https://github.com/kubernetes-sigs/kube-network-policies/blob/c7fc563db4e9f011d200da021b59098ac02e9bdb/pkg/networkpolicy/controller.go#L231

must be raised to 1280 (all extension headers _must_ fit into a minimum sized IPv6 packet).

uablrek:
I am unsure how fragments are handled. I don't think it's a problem since only the first packet of every connection is handled.

aojea:
do you want to take it Lars?

uablrek:
Yes. I would like to learn howto make bit manipulations in go (I have done it in C). The difficult thing is to check for too-short packets in every step.

/assign

aojea:
> must be raised to 1280 (all extension headers must fit into a minimum sized IPv6 packet).

yeah, I can see in multiple repos they initialize this value to `0xffff` so now I'm thinking I may did some premature optimization here  thinking that the less I copy the better, feel free to use a larger value you think is reasonable but please modify the stringer to not dump the entire packet , 

https://github.com/kubernetes-sigs/kube-network-policies/blob/c7fc563db4e9f011d200da021b59098ac02e9bdb/pkg/networkpolicy/packet.go#L23

just the first X bits ands is very useful for debugging  , special for DNS or plain protocols

```
I0425 16:32:52.606641       8 controller.go:364] Evaluating packet [44] 10.64.2.44:34220 10.0.186.101:80 TCP
00000000  00 00 a0 02 7f 94 d0 ff  00 00 02 04 05 8c 04 02  |................|
00000010  08 0a 40 e1 ed 7b 00 00  00 00 01 03 03 07        |..@..{........|
I0425 16:32:52.606688       8 controller.go:389] [Packet 44] Egress NetworkPolicies: 0 Allowed: true
[...]

PR #49, opened 2024-07-02
allow unknown protocols

aojea:
Unknown protocols are not subject to Kubernetes policies, hence they must be allowed


aojea:
Also check if the IPv6 failure in https://github.com/kubernetes-sigs/kube-network-policies/pull/47 are related to the output chain is something environmental

cc: @uablrek

aojea:
ok, the test passes here, so is the output rule 

/hold

I don't have clear what is the expected behavior https://github.com/kubernetes-sigs/network-policy-api/issues/187#issuecomment-2202722531

/assign @danwinship

pkg/networkpolicy/controller.go
```diff
@@ -656,6 +656,16 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
 			Name: chainName,
 		})
+
+		if hook == knftables.OutputHook {
+			// accept all outbound traffic to the loopback interface
+			tx.Add(&knftables.Rule{
+				Chain: chainName,
+				Rule: knftables.Concat(
+					"oif", "lo", "accept"),
+			})
```
aojea, line 666: @uablrek can you check in your PR this solves the problem?

uablrek: I added this update, but the address matches should never hit anything with "lo" since such packet would be a "martian"

uablrek: It worked! My world is crumbling! My reasoning is that since the set's only contains pod-addresses, packets matching whose sets can't be routed to/from "lo", since that would make them martians.

Unless... a SatefulSet pod actually have a loopback address as pod-address.

aojea: @astoycos @tssurya do you have an explanation for this failurs on the conformance suite only with ipv6?

uablrek: The modified chain for reference:
```
        chain output {
                type filter hook output priority filter - 5; policy accept;
                ct state established,related accept
                oif "lo" accept
                ip saddr @podips-v4 queue to 100 comment "process IPv4 traffic with network policy enforcement"
                ip daddr @podips-v4 queue to 100 comment "process IPv4 traffic with network policy enforcement"
                ip6 saddr @podips-v6 queue to 100 comment "process IPv6 traffic with network policy enforcement"
                ip6 daddr @podips-v6 queue to 100 comment "process IPv6 traffic with network policy enforcement"
        }
```

pkg/networkpolicy/controller.go
```diff
@@ -391,8 +391,8 @@ func (c *Controller) Run(ctx context.Context) error {
 
 		packet, err := parsePacket(*a.Payload)
 		if err != nil {
-			klog.Infof("Can not process packet %d applying default policy (failOpen: %v): %v", *a.PacketID, c.config.FailOpen, err)
-			c.nfq.SetVerdict(*a.PacketID, verdict) //nolint:errcheck
+			klog.Infof("Can not process packet %d, allowing ... : %v", *a.PacketID, err)
+			c.nfq.SetVerdict(*a.PacketID, nfqueue.NfAccept) //nolint:errcheck
```
danwinship, line 395: I'm not sure this is really correct

pkg/networkpolicy/packet.go
```diff
@@ -73,7 +73,7 @@ func parsePacket(b []byte) (packet, error) {
 		t.proto = v1.ProtocolSCTP
 		dataOffset = hdrlen + 8
 	default:
-		return t, fmt.Errorf("unknown protocol %d", protocol)
+		return t, fmt.Errorf("IP family %s unknown transport protocol %d", t.family, protocol)
```
danwinship, line 76: You don't want to log an error for every ICMPv6 message. You should add a `case 58:` for that.
Probably `parsePacket` needs another return value for "not an error, but just accept the packet".

(It seems that your callback only gets called for IP and IPv6 protocol packets, right? And, in particular, _not_ ARP and ICMP packets, which are separate L3 protocols, unlike ICMPv6/NDP, which are subprotocols of IPv6.)

aojea:
/hold cancel

@uablrek I've added you as co author and incorporated your commits with some amendments

pkg/networkpolicy/controller.go
aojea: this means this packet is not TCP or UDP or SCTP

pkg/networkpolicy/packet.go
aojea: see the next commit, only TCP , UDP and SCTP packets are sent to the userspace

pkg/networkpolicy/controller.go
```diff
@@ -665,6 +665,13 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 				"ct", "state", "established,related", "accept"),
 		})
 
+		// only process protocols supported by Kubernetes network policies
+		tx.Add(&knftables.Rule{
+			Chain: chainName,
+			Rule: knftables.Concat(
+				"meta", "l4proto", "!=", "{ tcp, udp, sctp }", "accept"),
```
aojea, line 682: ping @danwinship

pkg/networkpolicy/controller.go
aojea: it is not a flake, it is a test failure

danwinship: This rule definitely seems wrong.

I think sometimes packets are considered by the kernel to be traversing `lo` if they have the same local source and destination IP, even if that IP isn't 127.0.0.1/::1?

Anyway, the failure here is probably caused by the fact that in ipvs, packets traverse the network stack twice (from source to `ipvs0` and then from `ipvs0` to destination), and the DNAT and SNAT don't happen in exactly the same place as with iptables/nftables kube-proxy. (The DNAT happens inside `ipvs0`... I forget if the SNAT happens in the first or second traversal.)

Anyway, point is, the iptables/nftables logic is going to end up doing the wrong thing for some packets in ipvs mode. The fix _might_ be to ignore packets with `oif = "ipvs0"` so you only look at them on the second traversal, but I'm really not sure about that.

aojea: this rule fixes the IPv6 jobs, ipvs is unrelated here

pkg/networkpolicy/packet.go
uablrek: At the moment I think all fragments will also have an unknown protocol. But I am about to fix that (#15)

aojea:
So, thinking as an user, I think the best behavior will be that if I don't define any protocol in a network policy I apply the existing policy rules ... we can do this just not returning an error during packet processing , it will return an empty string for the protocol field and zero for src and destination port, and it will fail to match if there is a rule specifying a protocol

pkg/networkpolicy/controller.go
aojea: > I think sometimes packets are considered by the kernel to be traversing lo if they have the same local source and destination IP, even if that IP isn't 127.0.0.1/::1?

So, what I want to express is "we only need to process in OUTPUT local generated packets that are sent outside (not localhost dest)"

image

I had this issue in the past with kubelet probes and we solved it using the lo rule, I think it is correct but I can not explain the science behind it, just the evidence it works as expected

pkg/networkpolicy/controller.go
danwinship: No, it means "a parsing error occurred". I'm not sure we want to say all parse errors should be "accepts".

(I'm not sure we _don't_ want that either...)

aojea: I see, it can only happens if is not ip or ipv6 or one of those protocols ... but indeed is confusing the way it is now

uablrek: All packets with unknown protocol (not TCP or UDP or SCTP) _must_ be accepted. Unfortunately there is no `v1.ProtocolUnknown` constant, but we can type-cast:
```go
	default:
		//return t, fmt.Errorf("unknown protocol %d", protocol)
		t.proto = v1.Protocol("unknown")
```
but that feels a bit hackish...

aojea: we don't need to cast to v1.protocol type, we can use a string for that field or create our own enum
The code is this way just by commodity, but if it does not fit the design we should adapt it accordingly

pkg/networkpolicy/packet.go
uablrek: > see the next commit, only TCP , UDP and SCTP packets are sent to the userspace

That may fix the fragment problem. The kernel will reassemble packets before sending them to user-space (perhaps it already does in which case #15 isn't a kind/bug after all). Ref

pkg/networkpolicy/controller.go
uablrek: Would it be ok to just store the proto as an int in `struct packet`, and let the caller match on "syscall.IPPROTO_*". IMO that's the most flexible.

pkg/networkpolicy/controller.go
danwinship: > this rule fixes the IPv6 jobs, ipvs is unrelated here

oh... so... the issue is that if we add processing on the `output` hook, then IPv6 (but not IPv4) breaks unless we also add the `! lo` rule?

> So, what I want to express is "we only need to process in OUTPUT local generated packets that are sent outside (not localhost dest)"

Why should it matter though? We shouldn't be blocking any packets that are destined for localhost. With the `podips` changes, we shouldn't even intercept them...

> I can not explain the science behind it

boo!

> any packet destined to localhost does not have to be processed by network policies anyway, right?

A packet from a pod-network pod addressed to `127.0.0.0/8` or `::1` would not leave the pod netns, and it's not possible (by validation) to have an EndpointSlice that points to `127.0.0.0/8` or `::1`, so you can't end up with that via Service DNAT either. But you can add non-localhost IPs to `lo` I think?

pkg/networkpolicy/controller.go
danwinship: My point was that it doesn't seem right to say that every possible type of error from `parsePacket` should result in the packet being accepted. IMO it would be better to have `parsePacket` return both an error and a verdict, even if that verdict happens to be "accept" for all of the errors it _currently_ catches.

uablrek: After some investigation, I agree: all unrecognized packets shall not be accepted (https://github.com/kubernetes-sigs/kube-network-policies/pull/51#issuecomment-2216969955). However, I think `parsePacket()` should not set a verdict. It should parse the packet and let the caller decide, but then is can't return TCP,UDP,SCTP or an error for everything else.

aojea:
So, it seems there are two AI:
1. figure out why ipv6 fails https://github.com/kubernetes-sigs/kube-network-policies/pull/49#discussion_r1669094712
2. define the behavior on unrecognized packets https://github.com/kubernetes-sigs/kube-network-policies/pull/49#discussion_r1670034038

danwinship:
is there any part of this or #47 that can be merged as-is?

uablrek:
> is there any part of this or https://github.com/kubernetes-sigs/kube-network-policies/pull/47 that can be merged as-is?

IMO https://github.com/kubernetes-sigs/kube-network-policies/pull/47 can wait. We know the consequences (proxy-mode=ipvs doesn't work). I plan to return the IPPROTO number and reject some protocols, like packets with IPv6 routing headers, so I am unsure about this PR also.

uablrek:
I think accepting too much (all unknown) may be a security issue in the IPv6 extension header case. For instance if we miss some combination of EH's that are valid TCP, we may accept TCP traffic that should be blocked.

Better be restrictive, and for instance reject packets with EH's in a non-recommended order.

pkg/networkpolicy/controller.go
aojea: > Would it be ok to just store the proto as an int in `struct packet`, and let the caller match on "syscall.IPPROTO_*". IMO that's the most flexible.

that sounds like the best idea

uablrek: I have experimented some with this, and I think the best is to keep packet.proto as v1.Protocol, and add a new "ipproto int" in the packet structure.

An IPPROTO_ROUTING extension header, or extension header out-of-order will return a
```
var ErrorExtensionHeader = fmt.Errorf("unsupported extension header")
```

uablrek:
Scratch https://github.com/kubernetes-sigs/kube-network-policies/pull/49#issuecomment-2217717047.

Please merge this PR as-is an close #47. I'll rebase #51 if needed, then that PR should be next in line IMO. I have prepared a commit on top of #51 to handle EH's out-of-order, and routing-EH (https://github.com/kubernetes-sigs/kube-network-policies/pull/49#discussion_r1671898983)

danwinship:
OK, so thinking about this some more, if a pod is unprivileged and does not have `CAP_NET_ADMIN` or `CAP_NET_RAW`, then it can't send anything except IP. Given that, and the fact that we want to avoid _accidentally_ blocking ICMP reply packets (eg destination unreachable), I think we should not make any effort to ever block ICMP.

So, I'd say, we should keep the `meta l4proto != { tcp, udp, sctp } accept` rule and just not worry about anything else.

So that solves the "unrecognized protocol" problem and leaves:
- fragments, which I think we think isn't a problem if we have the `meta l4proto` rule? We should figure out how to test this...
- parse errors, which probably shouldn't happen, but I think we should treat as "deny", and log errors about, and see if they ever actually happen
- IPv6 extensions, which I believe we are intentionally punting to #51 and not worrying about in this patch

Right?

pkg/networkpolicy/controller.go
danwinship: (There's no real reason to use `knftables.Concat` here. You can just write it as a single string.)

aojea:
> OK, so thinking about this some more, if a pod is unprivileged and does not have `CAP_NET_ADMIN` or `CAP_NET_RAW`, then it can't send anything except IP. Given that, and the fact that we want to avoid _accidentally_ blocking ICMP reply packets (eg destination unreachable), I think we should not make any effort to ever block ICMP.
> 
> So, I'd say, we should keep the `meta l4proto != { tcp, udp, sctp } accept` rule and just not worry about anything else.
> 
> So that solves the "unrecognized protocol" problem and leaves:
> 
> * fragments, which I think we think isn't a problem if we have the `meta l4proto` rule? We should figure out how to test this...
> * parse errors, which probably shouldn't happen, but I think we should treat as "deny", and log errors about, and see if they ever actually happen
> * IPv6 extensions, which I believe we are intentionally punting to [Parse IPv6 extension headers #51](https://github.com/kubernetes-sigs/kube-network-policies/pull/51) and not worrying about in this patch
> 
> Right?

ICMP packets can be evaluated by network policies if no Port fields are set https://github.com/kubernetes-sigs/kube-network-policies/pull/51#discussion_r1681040192

I'm going to close this to avoid more noise, once we have the parser done, I want to debug the OUTPUT chain problem separetly

PR #50, opened 2024-07-03
Use structured logging

### LINKED ISSUE #48: Use structured logging
It looks nicer in code and output (if piped through [jq](https://jqlang.github.io/jq/)), and it's much better for post-processing.

This implies a more common-practice use of `context`. A "ctx" should be passed almost everywhere, not a "ctx.Done()".

I think I am almost as big fan of structured logging as @pohly  :smiley:, so I'll be happy to do a PR (but that will be a big one).

/kind feature

uablrek:
Use structured logging in json format if `-logging-format=json` is specified. The same setup functions as for K8s core packages are used ([component-base](https://github.com/kubernetes/component-base/tree/master)).

Also, [contextual logging](https://kubernetes.io/blog/2023/12/20/contextual-logging-in-kubernetes-1-29/) is used (beta in 1.30, so no feature-gateway is needed). 

Fixes #48 

This PR will be a draft until all code is changed to use structured/contextual logging.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -21,6 +21,9 @@ import (
 	v1 "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/component-base/logs"
+	logsapi "k8s.io/component-base/logs/api/v1"
+	_ "k8s.io/component-base/logs/json/register"
 	nodeutil "k8s.io/component-helpers/node/util"
 	"k8s.io/klog/v2"
 
@@ -52,17 +55,37 @@ func init() {
 	}
 }
 
+// This is a pattern to ensure that deferred functions executes before os.Exit
 func main() {
-	// enable logging
-	klog.InitFlags(nil)
+	os.Exit(run())
+}
+func run() int {
+	// Enable logging in the Kubernetes core package way (support json output)
+	// https://github.com/kubernetes/component-base/tree/master
+	c := logsapi.NewLoggingConfiguration()
+	logsapi.AddGoFlags(c, flag.CommandLine)
 	flag.Parse()
+	logs.InitLogs()
+	if err := logsapi.ValidateAndApply(c, nil); err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		return 1
+	}
+
+	// Create a context for structured logging, and catch termination signals
+	ctx, cancel := signal.NotifyContext(
+		context.Background(), os.Interrupt, unix.SIGINT)
+	defer cancel()
+
+	logger := klog.FromContext(ctx)
+	logger.Info("called", "args", flag.Args())
 
 	flag.VisitAll(func(flag *flag.Flag) {
-		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
+		logger.Info("flag", "name", flag.Name, "value", flag.Value)
 	})
 
 	if _, _, err := net.SplitHostPort(metricsBindAddress); err != nil {
-		klog.Fatalf("error parsing metrics bind address %s : %v", metricsBindAddress, err)
+		logger.Error(err, "parsing metrics bind address", "address", metricsBindAddress)
+		return 1
 	}
 
 	nodeName, err := nodeutil.GetHostname(hostnameOverride)
@@ -95,18 +118,6 @@ func main() {
 	if err != nil {
 		panic(err.Error())
 	}
-
-	// trap Ctrl+C and call cancel on the context
-	ctx := context.Background()
-	ctx, cancel := context.WithCancel(ctx)
-
-	// Enable signal handler
-	signalCh := make(chan os.Signal, 2)
-	defer func() {
-		close(signalCh)
-		cancel()
-	}()
-	signal.Notify(signalCh, os.Interrupt, unix.SIGINT)
 
 	informersFactory := informers.NewSharedInformerFactory(clientset, 0)
 
@@ -149,7 +160,8 @@ func main() {
 		cfg,
 	)
 	if err != nil {
-		klog.Fatalf("Can not start network policy controller: %v", err)
+		logger.Error(err, "Can not start network policy controller")
+		return 1
 	}
 	go func() {
 		err := networkPolicyController.Run(ctx)
@@ -161,13 +173,9 @@ func main() {
 		npaInformerFactory.Start(ctx.Done())
 	}
 
-	select {
-	case <-signalCh:
-		klog.Infof("Exiting: received signal")
-		cancel()
-	case <-ctx.Done():
-	}
+	<-ctx.Done()
 
 	// grace period to cleanup resources
 	time.Sleep(5 * time.Second)
+	return 0
 }
--- a/pkg/networkpolicy/controller.go
+++ b/pkg/networkpolicy/controller.go
@@ -140,7 +140,7 @@ func newController(client clientset.Interface,
 	broadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: client.CoreV1().Events("")})
 	recorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: controllerName})
 
-	klog.V(2).Infof("Creating controller: %#v", config)
+	klog.V(2).InfoS("Creating controller", "config", config)
 	c := &Controller{
 		client: client,
 		config: config,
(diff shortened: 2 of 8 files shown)


uablrek:
/cc @pohly

@pohly Can you please review the updates in `main()` in the first commit? I don't want to do any fundamental mistakes. The rest I think I can manage.

uablrek:
Example output:
```
# kubectl logs -n kube-system   kube-network-policies-8hb7p | jq
...
{
  "ts": 1720002972865.8306,
  "caller": "cache/reflector.go:359",
  "msg": "Caches populated for *v1.Pod from pkg/mod/k8s.io/client-go@v0.30.2/tools/cache/reflector.go:232",
  "v": 2
}
{
  "ts": 1720002972953.2607,
  "caller": "cache/shared_informer.go:320",
  "msg": "Caches are synced for kube-network-policies",
  "v": 0
}
{
  "ts": 1720002972953.3674,
  "caller": "networkpolicy/metrics.go:61",
  "msg": "Registering metrics",
  "v": 0
}
{
  "ts": 1720002972953.4724,
  "caller": "networkpolicy/controller.go:344",
  "msg": "Syncing nftables rules",
  "v": 0
}
```

uablrek:
Lint complaints:
```
Error: cmd/main.go:81:3: exitAfterDefer: os.Exit will exit, and `defer cancel()` will not run (gocritic)
```
This is just as bad with `klog.Fatalf`, but lint misses that. A pattern seem to be:
```go
func main() {
  os.Exit(run())
}
func run() int {
  return errorCode
}
```
Shall we use it? Else I restore `klog.Fatalf` in main()

aojea:
> ```go
> func main() {
>   os.Exit(run())
> }
> func run() int {
>   return errorCode
> }
> ```

that sounds like the right way to do it

aojea:
@uablrek what is missing here? it lgtms

/lgtm
/approve

uablrek:
> @uablrek what is missing here? it lgtms

Lots and lots of `Infof()`, and friends, should be changed to `InfoS()`.

Also i was planning to introduce contextual logging, but I think that can wait.

uablrek:
Structured logging allows a nice optimization possibility. Without this `p.String()` is always called, regardless of verbosity.

```go
	// evaluatePacket() should be fast unless trace logging is enabled.
	// Logging optimization: We check if V(2) is enabled before hand,
	// rather than evaluating the all parameters make an unnecessary logger call
	tlogger := logger.V(2)
	if tlogger.Enabled() {
		tlogger.Info("Evaluating packet", "packet", p)
		tlogger = tlogger.WithValues("id", p.id)
	}
```

pkg/networkpolicy/controller.go
```diff
@@ -371,7 +373,7 @@ func (c *Controller) Run(ctx context.Context) error {
 
 	nf, err := nfqueue.Open(&config)
 	if err != nil {
-		klog.Infof("could not open nfqueue socket: %v", err)
+		logger.Info("could not open nfqueue socket", "error", err)
```
uablrek, line 376: Logging like this exist at several places. I am unsure how to handle them. A general recommendation is to _not_ log errors if you return them (let the caller decide). And normally `logger.Error(err, ...)` is used for logging errors, but then it will be at a higher log-level. I made a compromise by keeping the log at Info level.

uablrek: A way to keep info about what caused the error is to [wrap](https://pkg.go.dev/errors) them:
```go
return fmt.Errorf("could not open nfqueue socket %w", err)
```
(ref: [recent discussion on slack](https://kubernetes.slack.com/archives/CHGFYJVAN/p1720715380174519))

uablrek:
Example packet log in JSON format:
```json
{
  "ts": 1720777151838.3552,
  "caller": "networkpolicy/controller.go:461",
  "msg": "Evaluating packet",
  "v": 2,
  "packet": {
    "Id": 9,
    "Family": "IPv4",
    "SrcIP": "11.0.3.3",
    "DstIP": "11.0.2.2",
    "Proto": "TCP",
    "SrcPort": 33803,
    "DstPort": 6000
  }
}
```

uablrek:
I will squash commits later

aojea:
let me know once is ready for merging

cmd/main.go
```diff
@@ -51,20 +54,40 @@ func init() {
...
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		return 1
+	}
 
-	klog.Infof("flags: %v", flag.Args())
+	// Create a context for structured logging, and catch termination signals
+	ctx, cancel := signal.NotifyContext(
+		context.Background(), os.Interrupt, unix.SIGINT)
+	defer cancel()
+	logger := klog.NewKlogr()
```
pohly, line 77: This shouldn't be needed and might never give you a JSON logger.

Leave the context as it is. A `klog.FromContext` will then get the global default prepared by `logsapi.ValidateAndApply`.

```suggestion
	logger := klog.FromContext(ctx)
```

cmd/main.go
```diff
@@ -51,20 +54,40 @@ func init() {
...
+	ctx = klog.NewContext(ctx, logger)
```
pohly, line 78: Delete this.

uablrek:
Rebased, updated after review, and use `klog.KObj`. Ready for review.

Please note that payload is empty for TCP/SYN, which is the only TCP packets we see:
```
I0721 08:56:53.060501       1 controller.go:491] "Evaluating packet" packet=<
        [10] 11.0.4.3:34751 11.0.4.2:6000 TCP
 >
```

uablrek:
> let me know once is ready for merging

@aojea Ready for review at least. I hope for merging. No logic is altered (except in main()), but very many log printouts

uablrek:
The flag-logging can be made in different ways:

1. The "structured object way" (this doesn't look so good for non-json logging):
```go
	flag.VisitAll(func(flag *flag.Flag) {
		logger.Info("flag", "flag", flag)
	})
```
```json
{
  "ts": 1731837080411.8074,
  "caller": "cmd/main.go:83",
  "msg": "flag",
  "v": 0,
  "flag": {
    "Name": "nfqueue-id",
    "Usage": "Number of the nfqueue used",
    "Value": 100,
    "DefValue": "100"
  }
}
```

2. The "name/value way" (this is what I selected)
```go
	flag.VisitAll(func(flag *flag.Flag) {
		logger.Info("flag", "name", flag.Name, "value", flag.Value)
	})
```
```json
{
  "ts": 1731837858630.011,
  "caller": "cmd/main.go:83",
  "msg": "flag",
  "v": 0,
  "name": "nfqueue-id",
  "value": "100"
}
```
Non-json logging:
```
I1117 11:05:07.914817  119614 main.go:83] "flag" name="nfqueue-id" value="100"
```

3. The "non structured logging way" (not recommended)
```go
	flag.VisitAll(func(flag *flag.Flag) {
		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
	})
```
```json
{
  "ts": 1731838042913.3442,
  "caller": "cmd/main.go:84",
  "msg": "FLAG: --nfqueue-id=\"100\"",
  "v": 0
}
```

uablrek:
The failing test seem unrelated to the updates. I ran e2e:
```
(BASE_FAMILY=IPv6 PROXY_MODE=nftables)
export SKIP=NOTHING
export FOCUS=Netpol
Ran 50 of 6622 Specs in 132.293 seconds
SUCCESS! -- 50 Passed | 0 Failed | 0 Pending | 6572 Skipped
```

aojea:
/lgtm
/hold

waiting for @paulgmiller review, we can unhold tomorrow if he does not have time to review

@paulgmiller just do `/hold cancel` if you are happy with the PR

pkg/networkpolicy/controller.go
```diff
@@ -444,11 +446,11 @@ func (c *Controller) Run(ctx context.Context) error {
 		}
 
 		startTime := time.Now()
-		klog.V(2).Infof("Processing sync for packet %d", *a.PacketID)
+		logger.V(2).Info("Processing sync for packet", "id", *a.PacketID)
```
paulgmiller, line 449: should we use the same tlogger := logger.V(2)  if tlogger.Enabled tricke we use here as we use at 517 in evaluage packet?

paulgmiller (approved):
Seems good sorry to be slow had to learn some things about k8s.io/component-base/logs/

pkg/networkpolicy/controller.go
uablrek: It's not as urgent in this case since only `PacketID` is passed (not an entire packet analyzed)

PR #51, opened 2024-07-05
Parse IPv6 extension headers

uablrek:
IPv6 extension headers must be parsed, at least fragment headers.

Fixes: #15


uablrek:
A unit test is added, that currently fails, but high-lite the fragmented IPv6-UPD problem.

uablrek:
The extension packets are _assumed_ to be in the [recommended](https://datatracker.ietf.org/doc/html/rfc2460#section-4.1), i.e. a fragment header is the last one, if it exist.

The IPPROTO_HOPOPTS, IPPROTO_ROUTING and IPPROTO_DSTOPTS extension headers are not tested. It is non-trivial to generate these header for testing.

uablrek:
The capture-chunk is still 128 byte. I think that the fragment case is the most important, and that doesn't require more

uablrek:
There may be problems also with IPv4 fragments. I think they will be handled as un-fragmented packets, but the L4-header is missing, so garbage data will be used. I will test it and rename this PR to something like "Parse IPv6 extension headers and handle fragments", if necessary.

uablrek:
FYI https://github.com/uablrek/pcap2go is used to create packet test-data

pkg/networkpolicy/packet.go
```diff
@@ -55,33 +57,51 @@ func parsePacket(b []byte) (packet, error) {
...
 		t.dstIP = make(net.IP, net.IPv6len)
 		copy(t.dstIP, b[24:40])
-		// NextHeader (not extension headers supported)
-		protocol = int(b[6])
+		// Handle extension headers. TODO: The recommended order is assumed!
+		nxtHeader = int(b[6])
+		l4offset = hdrlen
+		for nxtHeader == syscall.IPPROTO_DSTOPTS || nxtHeader == syscall.IPPROTO_HOPOPTS || nxtHeader == syscall.IPPROTO_ROUTING {
+			// These header have a lenght in 8-octet units, not including the first 8 octets
+			nxtHeader = int(b[l4offset])
```
aojea, line 65: we are capping the packet , should we need to validate `len(b)` to avoid a panic?

pkg/networkpolicy/packet_test.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+	}
+	if expected.dstPort > 0 {
+		if got.dstPort != expected.dstPort {
+			t.Fatalf("%s: dstPort: expected=%v, got=%v", tc, expected.dstPort, got.dstPort)
+		}
+	}
+}
+
+var packetsUdpFragIPv6 = [][]byte{
```
aojea, line 97: add a comment explaining how this is generated so future us can create more of these payloads

uablrek:
> There may be problems also with IPv4 fragments.

Probably not, because when conntrack is used (which it always is in K8s), [IPv4 packets are reassembled](https://unix.stackexchange.com/questions/650790/unwanted-defragmentation-of-forwarded-ipv4-packets).

pkg/networkpolicy/packet.go
uablrek: Yes, it should be checked in every step. I plan to make a unit-test that feeds capped packets from 1 byte to 128, and verify that no segv occurs.

uablrek:
Considering the above :point_up_2:, shall we assume the no IPv4 fragments are ever delivered to user-space, or should we be prepared to the very unlikely event?

aojea:
> Considering the above 👆, shall we assume that no IPv4 fragments are ever delivered to user-space, or should we be prepared for the very unlikely event?

seems reasonable, but leave a comment with the link to that great explanation to document the reason

pkg/networkpolicy/packet_test.go
uablrek: Comment added in packet_test.go, and pcap files added in test-data/

uablrek:
I added checks for packets with incomplete headers and corrupted packets (header lengths too small). But the kernel will verify this and discard those packets. If the capture length is changed 128->1280, all those checks should be redundant.

@aojea Do you think I should set capture length 1280, and remove all too-small and corrupted tests?

uablrek:
IPPROTO_DSTOPTS may appear again after IPPROTO_FRAGMENT in the recommended order. I will fix that when I find a way to test it.

pkg/networkpolicy/packet_test.go
```diff
@@ -0,0 +1,231 @@
...
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+// captureChunk is the number of bytes transfered to user-space
+// in the nfqueue
+const captureChunk = 128
+
+func TestUdpFragmentIPv6(t *testing.T) {
```
danwinship, line 19: "UDP" not "Udp". Golang style doesn't force acronyms to "titlecase" (though they can be all-lowercase if needed).

(Likewise everywhere.)

pkg/networkpolicy/packet_test.go
```diff
@@ -0,0 +1,231 @@
...
+	// 3. Identified as an UDP packet
+	tests := []struct {
+		name     string
+		input    []byte
+		err      bool
+		expected packet
+	}{
+		{
+			name:  "UDP first fragment",
+			input: packetsUdpFragIPv6[0][14:(14 + captureChunk)],
```
danwinship, line 36: `14` should not be a magic number

pkg/networkpolicy/packet_test.go
danwinship, line 233: seems like this would make more sense as three separate variables with useful names.
also, maybe split into L2 header, L3 header, payload, with some minimal comments?

pkg/networkpolicy/packet_test.go
```diff
@@ -64,6 +69,136 @@ func TestUdpFragmentIPv6(t *testing.T) {
...
+			expected: packet{
+				family:  v1.IPv4Protocol,
+				proto:   v1.ProtocolTCP,
+				dstIP:   net.ParseIP("192.168.1.1"),
+				dstPort: 5001,
+				srcIP:   net.ParseIP("192.168.1.2+1"),
+			},
+		},
+		{
+			name:  "PSH, ACK",
```
danwinship, line 114: we should only be receiving packets for `ctstate new` connections...

pkg/networkpolicy/packet_test.go
uablrek: Removed

pkg/networkpolicy/packet_test.go
uablrek: I saved the original pcap files in `test-data/` so people can analyze it with `wireshark` for instance

pkg/networkpolicy/packet_test.go
uablrek: Yeah, but IMHO `parsePacket()` should be more-or-less generic. It should not panic if some TCP packet arrives that is not a SYN for instance. I think "test too much" here should be ok.

uablrek:
TL;DR; SRv6 is a horrible edge use-case!

In my quest for IPv6 packets with extension headers I revisited SRv6. I tested it after viewing an [excellent presentation](https://www.youtube.com/watch?v=vJaOKGWiyvU&list=PLj6h78yzYM2P4FvE6vARKAUg9BUi7ydw8&index=4) by Daniel Bernier, Bell Canada at KubeCon in Valencia 2022.

With SRv6 there is a routing header, followed by _another_ IPv6 header, and _then_ the TCP header:

tcp-syn-segment-routing

This will probably never happen, but network-policies _is_ in the forwarding plane, so it _might_ be possible so see something like this.

As I see it we have two choices:

1. Parse this and return addresses from the second IPv6 header, and detect proto TCP
2. Reject as "not supported"

The second can be "reject any packet with a routing header".

uablrek:
BTW, note the `TCP Option - Maximum segment size: 1440 bytes`. With SRv6 PMTU discovery doesn't work(!).

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +30,113 @@ func (p packet) String() string {
 // https://github.com/golang/net/blob/master/ipv4/header.go
 func parsePacket(b []byte) (packet, error) {
 	t := packet{}
-	if b == nil {
-		return t, fmt.Errorf("empty payload")
+	var errorTooShort = fmt.Errorf("Packet too short")
```
aojea, line 33:
```suggestion
	var errorTooShort = fmt.Errorf("packet too short")
```

linter complains 

> ST1005: error strings should not be capitalized (stylecheck)

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +30,113 @@ func (p packet) String() string {
...
+	var errorCorrupted = fmt.Errorf("Packet corrupted")
```
aojea, line 34:
```suggestion
	var errorCorrupted = fmt.Errorf("packet corrupted")
```

ditto

aojea: I think is better if you define the variables as global in the package so the consumer can cast the errors based on the error type

pkg/networkpolicy/packet_test.go
aojea: better add it under the same folder `pkg/networkpolicy/test-data/` and add a small README.md doc in the folder explaining https://github.com/uablrek/pcap2go

aojea:
> @aojea Do you think I should set capture length 1280, and remove all too-small and corrupted tests?

I prefer to make the code more resilient than depending on the kernel to make the right thing

There are some linter errors and some job failing that I do not know if is a flake, but this should be ready for review

uablrek:
I made the PR "ready for review". IPv6 fragments are handled, but there are some "exotic" cases that should also be covered, like [SRv6](https://github.com/kubernetes-sigs/kube-network-policies/pull/51#issuecomment-2216969955), and [repeated destopts](https://github.com/kubernetes-sigs/kube-network-policies/pull/51#issuecomment-2212477009). Perhaps extension headers that are _not_ in the recommended order, but I rather reject those. Also unit-tests for extension headers (except fragments) are missing.

But since IPv6 fragment handling is most urgent, and may be considered a bug, the remaining stuff can be added in a later kind/feature PR. I'll be happy to continue working with that.

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +33,110 @@ func (p packet) String() string {
...
+			}
+		}
+		protocol = nxtHeader
 	default:
-		return t, fmt.Errorf("unknown versions %d", version)
+		return t, fmt.Errorf("unknown version %d", version)
 	}
 
-	var dataOffset int
+	havePorts := true
```
aojea, line 109: when is this false? preparing for icmp?

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +33,110 @@ func (p packet) String() string {
...
+				nxtHeader = int(b[l4offset])
+				l4offset += 8
+				// Here it's assumed that the fragment is the last
+				// extension header before the L4 header. But this is
+				// only a *recommendation*.
+				// TODO: handle extension headers in non-recommended order
+			} else {
+				// Not first fragment, we have no L4 header and the
+				// payload begins after this header
+				l4offset = -1
```
aojea, line 101: is `protocol = nxtHeader` correct in this case? it will be `syscall.IPPROTO_FRAGMENT ` 

`l4offset = -1`. does not seem to be handled later, should this return here?

uablrek: > is protocol = nxtHeader correct in this case? it will be syscall.IPPROTO_FRAGMENT

It's intentional, but strictly not correct since there is an L4 protocol in the nxtHeader field. _But there is no L4 header!_ The fragment data begins immediately after the fragment header. So to identify the packet as a fragment, syscall.IPPROTO_FRAGMENT is kept in the `protocol` variable. I'll make a comment.

uablrek: > l4offset = -1. does not seem to be handled later, should this return here?

It's probably a just-in-case left-over. I'll check it.

uablrek:
Updated after comments.

I figured that we can return an empty `packet.proto` for fragments, and for any other packets we want to accept. The default is however still to return an error for non tcp/udp/sctp. I was uncertain about this, since it _requires_ that only tcp/udp/sctp packets are routed to the nfqueue.

The alternative is to return an empty `packet.proto` for _all_ packets we want to accept, not only fragments. If we do that, no "allow unknown protocols" pr is needed. The control is in `parsePacket()`.

I have prepared a commit that rejects IPPROTO_ROUTING headers, and handle the remaining case for headers in the recommended order: extra IPPROTO_DSTOPTS headers after the first fragment header. But it needs rebase after these updates. I also add a `ipproto` field in the packet struct. Not really needed, but "good to have" for logging and future extensions.

pkg/networkpolicy/packet.go
uablrek: removed

uablrek:
I am still working on tests, both unit-test and real traffic tests, with IPv6 extension headers. I have not figured out an easy way, but am planning to use a [tap device](https://en.wikipedia.org/wiki/TUN/TAP) that adds some IPv6 extension headers. I.e. packets are routed to a tap device, a user-space program add some headers, and send the packets back. I have done it before, and got quite far. IMO tap devices, same as nfqueue, are often overlooked for feature implementations and testing. A hyped alternative is eBPF, but that is much more complicated, and should be avoided unless performance is _really_ needed.

A good example of the power of tap devices is [tayga](http://www.litech.org/tayga/) NAT64 (rock-solid since 2011!)

uablrek:
I discovered that lksctp (Linux kernel SCTP) doesn't fragment on IP-level (this was new to me). The MTU of the outgoing device is examined, and fragmentation is made on SCTP level. This means that IPv6 fragments will never be used for lksctp. I don't know about other sctp implementations, but I will only test lksctp.

pkg/networkpolicy/controller.go
```diff
@@ -425,6 +425,15 @@ func (c *Controller) Run(ctx context.Context) error {
...
 			return 0
 		}
+		switch packet.proto {
+		case v1.ProtocolUDP:
+		case v1.ProtocolTCP:
+		case v1.ProtocolSCTP:
+		default:
+			// Accept packets that are parsed OK, but not TCP/UDP/SCTP.
+			c.nfq.SetVerdict(*a.PacketID, nfqueue.NfAccept) //nolint:errcheck
+			return 0
```
aojea, line 435: Dan and I were discussed offline, and we concluded that:
1. network policies only care about L4 protocol if you set the Ports fields
2. if not Port field is specified , is legit to assume that the user/admin can consider the network policy as "any" protocol (some from sftim about users wanting to control ICMP is spot on)

As conclusion, I think we set the packet.proto to empty , then srcPort and dstPort will be 0, and IIUIC the network policy logic will be able do the right thing.

This leaves the `parsePacket()` errors to short or malformed packets (module bugs) than we can say we deny by default , as we don't expect at this point to receive a packet from the kernel that is malformed

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +33,111 @@ func (p packet) String() string {
...
-	case 132:
+		payloadOffset = l4offset + 8
+	case syscall.IPPROTO_SCTP:
 		t.proto = v1.ProtocolSCTP
-		dataOffset = hdrlen + 8
+		payloadOffset = l4offset + 8
 	default:
+		// This "should not happen" since only { tcp, udp, sctp } are
+		// routed to the nfqueue
 		return t, fmt.Errorf("unknown protocol %d", protocol)
```
aojea, line 127: we are going to route any protocol here, so we can emit verdict on icmp per example, more on this in https://github.com/kubernetes-sigs/kube-network-policies/pull/51/files#r1681040192

pkg/networkpolicy/controller.go
aojea: can you just remove this switch case?

uablrek: And just check for an empty v1.Protocol? What is the best if-condition:
```go
  if string(packet.proto) == ""
```
?

uablrek: (I couldn't figure that out, hence the switch)

aojea: oh, not check at all , just pass the packet as is to the next stage

aojea: `c.evaluatePacket(packet)` will evaluate the packet only with the IPs information

uablrek: pushed

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +33,111 @@ func (p packet) String() string {
...
+		return t, nil
+
 	}
-	// TCP, UDP and SCTP srcPort and dstPort are the first 4 bytes after the IP header
-	t.srcPort = int(binary.BigEndian.Uint16(b[hdrlen : hdrlen+2]))
-	t.dstPort = int(binary.BigEndian.Uint16(b[hdrlen+2 : hdrlen+4]))
+	if payloadOffset > len(b) {
+		// We must have a complete L4 header, however payload may be zero-lengh
+		return t, ErrorTooShort
+	}
```
aojea, line 132: is this invalid? why this should be an error ?

aojea: ah, ok, if payloadOffset is equal to len(b) is a valid packet

pkg/networkpolicy/packet.go
```diff
@@ -29,59 +33,111 @@ func (p packet) String() string {
...
+	t.srcPort = int(binary.BigEndian.Uint16(b[l4offset : l4offset+2]))
+	t.dstPort = int(binary.BigEndian.Uint16(b[l4offset+2 : l4offset+4]))
+
 	// Obtain the offset of the payload
 	// TODO allow to filter by the payload
-	if len(b) >= hdrlen+dataOffset {
-		t.payload = b[hdrlen+dataOffset:]
+	if payloadOffset > 0 {
```
aojea, line 138: can this happen? I don't know if I prefer the old check

uablrek: dataOffset is a tcp-header thing. It should be used to compute the hdrlen for tcp-headers only. I think there was a bug in the old code for sctp and udp payloads. I looked that the hdrlen was added twice, since it was included in dataOffset (but I didn't really check carefully, since I re-wrote it)

uablrek: Oh, payloadOffset must be >0 now since we return in the default: branch. I can remove the check.

aojea:
/lgtm
/approve 

Thanks

PR #53, opened 2024-07-16
move nftables creation inside the controller constructor

aojea:
/assign @danwinship

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"sigs.k8s.io/knftables"
 	"sigs.k8s.io/kube-network-policies/pkg/networkpolicy"
 	npaclient "sigs.k8s.io/network-policy-api/pkg/client/clientset/versioned"
 	npainformers "sigs.k8s.io/network-policy-api/pkg/client/informers/externalversions"
@@ -57,11 +56,6 @@ func main() {
 	flag.Parse()
 
 	klog.Infof("flags: %v", flag.Args())
-
-	nft, err := knftables.New(knftables.InetFamily, "kube-network-policies")
-	if err != nil {
-		klog.Fatalf("Error initializing nftables: %v", err)
-	}
 
 	if _, _, err := net.SplitHostPort(metricsBindAddress); err != nil {
 		klog.Fatalf("error parsing metrics bind address %s : %v", metricsBindAddress, err)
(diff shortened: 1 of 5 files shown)


cmd/main.go
```diff
@@ -150,6 +143,9 @@ func main() {
 		banpInformer,
 		cfg,
 	)
+	if err != nil {
+		panic(err.Error())
+	}
```
danwinship, line 148: Why panic when it was `klog.Fatalf` before?
Semantically, this doesn't seem panic-worthy

pkg/networkpolicy/controller_test.go
```diff
@@ -92,27 +101,201 @@ func newController() *networkpolicyController {
 	npaClient := npaclientfake.NewSimpleClientset()
 	npaInformerFactory := npainformers.NewSharedInformerFactory(npaClient, 0)
 
-	controller := NewController(client,
-		&knftables.Fake{},
```
danwinship, line 96: Ah, that's the answer to "why is the knftables.Interface an argument".
(FTR, `&knftables.Fake{}` is wrong; you should use `knftables.NewFake()`)

pkg/networkpolicy/controller_test.go
```diff
@@ -92,27 +101,201 @@ func newController() *networkpolicyController {
...
+	}
+
+	networkpolicyInformer := informersFactory.Networking().V1().NetworkPolicies()
+	namespaceInformer := informersFactory.Core().V1().Namespaces()
+	podInformer := informersFactory.Core().V1().Pods()
+	nodeInformer := informersFactory.Core().V1().Nodes()
+	adminNetworkPolicyInformer := npaInformerFactory.Policy().V1alpha1().AdminNetworkPolicies()
+
+	baselineAdminNetworkPolicyInformer := npaInformerFactory.Policy().V1alpha1().BaselineAdminNetworkPolicies()
+	c := &Controller{
```
danwinship, line 116: blank line in the wrong place?

pkg/networkpolicy/controller_test.go
```diff
@@ -92,27 +101,201 @@ func newController() *networkpolicyController {
...
+			if pod.Spec.HostNetwork {
+				return []string{}, nil
+			}
+			result := []string{}
+			for _, ip := range pod.Status.PodIPs {
+				result = append(result, string(ip.IP))
+			}
+			return result, nil
+		},
+	})
```
danwinship, line 141: it seems crazy to duplicate all of this in the unit test; you should just have a `newControllerInternal()` that takes the knftables arg.

or call `NewController()` and just overwrite `c.nft` afterward

cmd/main.go
aojea: yeah it has to exit with error because the controller is not able to start

PR #55, opened 2024-07-19
Reject unsupported IPv6 extension headers

uablrek:
IPv6 extensions headers _must_ be in the [recommended order](https://datatracker.ietf.org/doc/html/rfc8200#section-4.1).

Routing headers are not supported, please see https://github.com/kubernetes-sigs/kube-network-policies/pull/51#issuecomment-2216969955.

For non-first fragments, the `ipproto` is set to `syscall.IPPROTO_FRAGMENT`, since it's important to know that the packet is a fragment, and _not_ important what it's a fragment of (e.g. UDP).

Packets with security headers (AH and ESP) are not rejected. Perhaps they should(?).

An 'ipproto' field is added in the packet struct. This field is currently not used, but will be needed for future extensions, like handling of ICMP (and ICMP6) packets. It may also be useful for logging (like the "payload" field).


uablrek:
Set as "draft" until unit-tests are added.

uablrek:
Finally! Packets with a `DSTOPTS` extension header (EH):

udp-dstopt

I wasted hours of trouble-shooting before discovering that the kernel silently discard packets with a DSTOPTS EH with _only_ "Pad" fields :angry: I had to make up an unknown option.

I plan to add another unit-test with `DSTOPTS` both after the IPv6 header _and_ after the fragment header. It is hard to generate packets for error tests, since the kernel would drop them.

uablrek:
> kernel silently discard packets with a DSTOPTS EH with only "Pad" fields

Not entirely true. Linux discard packets with "unnecessary" pad's. The RFC says strictly that Pad's should only be used for 8-byte alignment. I now can use minimum length (8 byte) EH's without problems, and option type 0x1e (experimental) is used for other types.

Here is an article that tells everything you _ever_ want to know about IPv6 extension headers on the internet (and then some):

https://blog.apnic.net/2022/10/13/ipv6-extension-headers-revisited/

**TL;DR;** IPv6 extension headers are best avoided in the public IPv6 network

(not even fragmentation headers are handled ok)

Anyway, DSTOPTS before _and_ after a fragmentation header is now unit-tested. However, Linux doesn't support them, so it may be an overkill.

Issue #65, opened 2024-07-23
TCP liveness/readiness probes are blocked by network policies


aojea:
> I added TCP liveness/readiness probes, and if I apply my "deny ingress" rule, the probes fail. This make sense to me, a deny ingress should block the probes. @aojea Do you know if there is some "don't do that" for this situation documented somewhere. Please, don't search for it, I can do that myself, but if you happen to know...

@danwinship is the source of truth for network policies 😄  and has also revisited the problem of probes recently, it also mentions this in the KEP https://github.com/kubernetes/enhancements/pull/4558
```
Changing the definition of NetworkPolicy v1 to remove the "kubelet
  can access all local pods" hole, even in the case where the cluster
  is configured in a way to forbid the use of probes that require the
  hole. Even if _kubelet_ no longer needs to access pods, some users
  likely depend on that rule to allow access to some pods from other
  host-level processes, and there is no other way in NetworkPolicy v1
  to write such a rule if we remove the built-in one
```

I think that traffic originated from the node should not be subject for network policies.

I checked with nftables trace and all trafic coming from the node has the `iif` `lo` independently that the destination IP belongs to another interface
```
trace id 2da4b9ff ip filter INPUT packet: iif "lo" @ll,0,112 0x800 ip saddr 192.168.8.2 ip daddr 192.168.8.2 ip dscp cs0 ip ecn not-ect ip ttl 64 ip id 30324 ip length 60 tcp sport 41872 tcp dport 10250 tcp flags == syn tcp window 33280
```

_Originally posted by @aojea in https://github.com/kubernetes-sigs/kube-network-policies/issues/54#issuecomment-2245178257_

uablrek:
I will probably not work with the implementation, but I am happy to help with testing and review

aojea:
The trace is in INPUT and the traffic directed to the Pods will not come from that hook, the rule has to be in POSTROUTING

This is an example trace
```
trace id a14a3f32 ip filter trace_chain2 packet: oif "veth33dfd85a" ip saddr 10.244.2.1 ip daddr 10.244.2.5 ip dscp cs0 ip ecn not-ect ip ttl 64 ip id 18830 ip length 60 tcp sport 55132 tcp dport 80 tcp flags == syn tcp window 32120
trace id a14a3f32 ip filter trace_chain2 rule tcp dport 80 meta nftrace set 1 (verdict continue)
trace id a14a3f32 ip filter trace_chain2 verdict continue
[...]

PR #66, opened 2024-07-23
node local traffic is not processed by network policies

aojea:
Fixes: https://github.com/kubernetes-sigs/kube-network-policies/issues/65

### DIFF
--- a/hack/ci/e2e.sh
+++ b/hack/ci/e2e.sh
@@ -168,8 +168,9 @@ run_tests() {
   fi
 
   # ginkgo regexes
-  SKIP="${SKIP:-"Feature|Federation|PerformanceDNS|DualStack|Disruptive|Serial|KubeProxy|GCE|Netpol|NetworkPolicy|256.search.list.characters|LoadBalancer.Service.without.NodePort|type.and.ports.of.a.TCP.service|loadbalancer.source.ranges"}"
-  FOCUS="${FOCUS:-"\\[sig-network\\]"}"
+  SKIP="${SKIP:-}"
+  FOCUS="${FOCUS:-}"
+  LABEL_FILTER="${LABEL_FILTER:-"Feature: containsAny NetworkPolicy"}"
   # if we set PARALLEL=true, skip serial tests set --ginkgo-parallel
   if [ "${PARALLEL:-false}" = "true" ]; then
     export GINKGO_PARALLEL=y
--- a/pkg/networkpolicy/controller.go
+++ b/pkg/networkpolicy/controller.go
@@ -695,6 +695,13 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 			Rule: knftables.Concat(
 				"icmpv6", "type", "{", "nd-neighbor-solicit, nd-neighbor-advert", "}", "accept"),
 		})
+		// Don't process traffic generated from the root user in the Node, it can block kubelet probes
+		// or system daemons that depend on the internal node traffic to not be blocked.
+		// Ref: https://github.com/kubernetes-sigs/kube-network-policies/issues/65
+		tx.Add(&knftables.Rule{
+			Chain: chainName,
+			Rule:  "meta skuid 0 accept",
+		})
 		// instead of aggregating all the expresion in one rule, use two different
 		// rules to understand if is causing issues with UDP packets with the same
 		// tuple (https://github.com/kubernetes-sigs/kube-network-policies/issues/12)


aojea:
/assign @danwinship 
/cc @uablrek @vaskozl 

Dan please check this is aligned with admin network policies also

pkg/networkpolicy/controller.go
```diff
@@ -695,6 +695,14 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 			Rule: knftables.Concat(
 				"icmpv6", "type", "{", "nd-neighbor-solicit, nd-neighbor-advert", "}", "accept"),
 		})
+		// Don't process traffic generated from the Node, as it can block kubelet probes
+		// or system daemons that depend on the internal node traffic to not be blocked.
+		// Ref: https://github.com/kubernetes-sigs/kube-network-policies/issues/65
+		tx.Add(&knftables.Rule{
+			Chain: chainName,
+			Rule: knftables.Concat(
+				"fib", "saddr", "type", "local", "accept"),
```
danwinship, line 704: (nit: you don't need to use `knftables.Concat` here; you can just write `"fib saddr type local accept"`)

But that's not right; with some network plugins, all local pod IPs would be considered "type local" (because the IPs are visible on interfaces in the host netns).

I think the "standard" approach is to allow only from the primary and secondary node IP (learned either via the downward API, or by watching the Node object). Though actually, I don't think that's CNI-agnostic; in some cases, node-to-pod traffic will use the IP of the bridge that the pods are on (eg `docker0`) rather than the official node IP.

The super-clever option would be to figure out kubelet's cgroup ID and write a `meta cgroupv2 ...` rule. Although I think some people expect the non-clever "allow from the node's IP" version anyway.

danwinship:
> Dan please check this is aligned with admin network policies also

ANP v1alpha1 does not provide any way to block node-to-pod traffic, so this doesn't affect ANP at all

pkg/networkpolicy/controller.go
danwinship: what was wrong with the `iif lo` approach?

hack/ci/e2e.sh
```diff
@@ -168,8 +168,6 @@ run_tests() {
   fi
 
   # ginkgo regexes
-  SKIP="${SKIP:-"Feature|Federation|PerformanceDNS|DualStack|Disruptive|Serial|KubeProxy|GCE|Netpol|NetworkPolicy|256.search.list.characters|LoadBalancer.Service.without.NodePort|type.and.ports.of.a.TCP.service|loadbalancer.source.ranges"}"
-  FOCUS="${FOCUS:-"\\[sig-network\\]"}"
```
danwinship, line 172: why?

aojea: we have now LABEL_FILTER that is the new cool kid on the block https://github.com/kubernetes/test-infra/pull/33075 ... you even can read the filter and understand it 😄

pkg/networkpolicy/controller.go
aojea: We have contributors willing to accept challenges , looking at @uablrek now :) 

I like the idea of restricting the loophole to the kubelet, now that you explain it the "type local" sounds easy to cheat

aojea: > what was wrong with the iif lo approach?

the trace does not show the iif metadata on the postrouting hook, so I think that to make that work we should add a new rule in INPUT to mark the packets and then in POSTROUTING ignore those packets

> cgroups

```
cat /proc/66685/cgroup
0::/kubelet.slice/kubelet.service
```

I don't know how stable or configurable is this option, @BenTheElder do you know?

There is also a skuid and skgid match
https://wiki.nftables.org/wiki-nftables/index.php/Matching_packet_metainformation

```
cat /proc/66685/status
Name:   kubelet
Umask:  0022
State:  S (sleeping)
Tgid:   66685
Ngid:   0
Pid:    66685
PPid:   1
TracerPid:      0
Uid:    0       0       0       0
Gid:    0       0       0       0
```

Maybe accepting everything that comes as root from the host is the best approach, kubelet hast to run as root and if you are root does not matter network policies, you can do whatever you want and access the pods directly on the namespaces

```
        chain trace_chain2 {
                type filter hook postrouting priority filter - 1; policy accept;
                meta skuid 0 counter packets 194 bytes 20003

```

uablrek:
Applied this PR and tested:
```
Versions:
  16:03:02 Server Version: v1.30.3
  16:03:02 CNI-plugin; flannel:v0.24.2
  16:03:02 Proxy-mode: "nftables"
  16:03:02 crio version 1.29.2
  16:03:02 BaseFamily=IPv6
E2e:
export SKIP=NOTHING
export FOCUS=Netpol
Ran 50 of 6652 Specs in 145.227 seconds
SUCCESS! -- 50 Passed | 0 Failed | 0 Pending | 6602 Skipped
```
And my PODs with TCP probes stays alive.

pkg/networkpolicy/controller.go
uablrek: I think you should go back to "type local". It's probably not "more unsafe" than anything existing, and to assume user 0 for all traffic I think will cause problems. People try very hard to _not_ enforce the root user it seems.

aojea:
pushed new version using the uid of root to discriminate the local packets

pkg/networkpolicy/controller.go
uablrek: Ok, maybe not "type local", but I think it's unsafe to assume that kubelet runs as root in all clusters, and will so forever

uablrek: But then again, I have no better idea.

aojea: kubelet has to do root operations creating cgroups per example, the probes need to open sockets ... it can be rootless https://kubernetes.io/docs/tasks/administer-cluster/kubelet-in-userns/ but in a rootless environment this project will not work too 😄 , the network there is 🕺 


> The network namespace of the Node components has to have a non-loopback interface, which can be for example configured with [slirp4netns](https://github.com/rootless-containers/slirp4netns), [VPNKit](https://github.com/moby/vpnkit), or [lxc-user-nic(1)](https://www.man7.org/linux/man-pages/man1/lxc-user-nic.1.html).
> 
> The network namespaces of the Pods can be configured with regular CNI plugins. For multi-node networking, Flannel (VXLAN, 8472/UDP) is known to work.
> 
> Ports such as the kubelet port (10250/TCP) and NodePort service ports have to be exposed from the Node network namespace to the host with an external port forwarder, such as RootlessKit, slirp4netns, or [socat(1)](https://linux.die.net/man/1/socat).

@BenTheElder  how safe do you think it is assume that kubelet runs as root?

uablrek:
Tested the new version. Same result as https://github.com/kubernetes-sigs/kube-network-policies/pull/66#issuecomment-2245359713 :+1:

PR #71, opened 2024-08-06
use different image for building

aojea:


cloudbuild.yaml
```diff
@@ -3,6 +3,6 @@ options:
   substitution_option: ALLOW_LOOSE
   machineType: E2_HIGHCPU_32
 steps:
-- name: gcr.io/k8s-testimages/krte:latest-master
+- name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud
```
ameukam, line 6:
```suggestion
- name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20240718-5ef92b5c36
```

Maybe stick to a tag instead of latest ?

Issue #70, opened 2024-08-06
Failing postsubmit job for building images


aojea:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kube-network-policies-image/1820461457674866688

```
docker buildx build . \
	--platform="linux/amd64,linux/arm64" \
	--tag="gcr.io/k8s-staging-networking/kube-network-policies:v20240805-4c4d7de" \
	--push
error: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
make: *** [Makefile:46: image-push] Error 1
ERROR
ERROR: build step 0 "gcr.io/k8s-testimages/krte:latest-master" failed: step exited with non-zero status: 2
--------------------------------------------------------------------------------
2024/08/05 14:10:35 Failed to run some build jobs: [error running [gcloud builds submit --verbosity info --config /home/prow/go/src/github.com/kubernetes-sigs/kube-network-policies/cloudbuild.yaml --substitutions _PULL_BASE_SHA=4c4d7de2845e4ff6ccf221b2ca3fdef940567c26,_GIT_TAG=v20240805-v0.5.1-5-g4c4d7de --project k8s-staging-networking --gcs-log-dir gs://k8s-staging-networking-gcb/logs --gcs-source-staging-dir gs://k8s-staging-networking-gcb/source gs://k8s-staging-networking-gcb/source/5b41ecca-d695-4ef1-9eb1-149c0d55fdaa.tgz]: exit status 1]

```

/help

@liangyuanpeng in case you are interested
cc @ameukam

ameukam:
Look like we should use `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud` instead of `krte` of the build image.

https://github.com/kubernetes-sigs/kube-network-policies/blob/2ee42ee98d44f12b37800437181779cdd31d7391/cloudbuild.yaml#L6-L8

BenTheElder:
This is passing, right?

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kube-network-policies-image/1889063786787115008

aojea:
> This is passing, right?
> 
> https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-kube-network-policies-image/1889063786787115008

it seems so https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-kube-network-policies-image

PR #72, opened 2024-08-07
chore: using self docker builder for multi-platform images.

liangyuanpeng:
releated https://github.com/kubernetes-sigs/kube-network-policies/issues/70

```
error: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")
```

following the error log to switch other builder for build image.

I have check other project for multi-platform images and basically, a specific builder will be created.

- https://github.com/kubernetes/ingress-nginx/blob/main/hack/init-buildx.sh
- https://github.com/kubernetes-sigs/kwok/blob/da3756aca50a589d084592ee0ded8ffd425e566e/cloudbuild.yaml#L24

and the file `hack/init-buildx.sh` is copy from https://github.com/kubernetes-sigs/kind/blob/main/hack/build/init-buildx.sh

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -3,6 +3,6 @@ options:
   substitution_option: ALLOW_LOOSE
   machineType: E2_HIGHCPU_32
 steps:
-- name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud
+- name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20240718-5ef92b5c36
   entrypoint: make
-  args: ['image-push']
+  args: ['release']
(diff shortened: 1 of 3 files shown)


cloudbuild.yaml
```diff
@@ -3,6 +3,6 @@ options:
   substitution_option: ALLOW_LOOSE
   machineType: E2_HIGHCPU_32
 steps:
-- name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud
+- name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20240718-5ef92b5c36
```
liangyuanpeng, line 6: stick to a tag  https://github.com/kubernetes-sigs/kube-network-policies/pull/71#discussion_r1705670459

/assign @aojea @ameukam

aojea:
/lgtm
/approve

let's try

Thanks

aojea:
@liangyuanpeng it seems is not working fine https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kube-network-policies-image/1821313751584870400

PR #94, opened 2024-10-03
performance: optimize memory usage

howardjohn:
In a ~15k Pod cluster, I see this drop memory footprint from about 200MB to 66MB.

This has a few improvements:
* Don't just strip managed fields, but keep only exactly what we want.
* Intern strips to reduce duplication of common things (label keys are almost always duplicated, etc)
* During startup only, GC more aggressively. The problem with the above approach is we get the full object *then* drop it. So if we don't manually GC, we will bloat up and not recover for a long time.

Some of this is overkill I think, I will see what I can remove and see what benefits remain so we don't have too much complexity where we don't get benefits.

Before/after: see the green vs purple line, ignore the rest...

### DIFF
--- a/pkg/networkpolicy/controller.go
+++ b/pkg/networkpolicy/controller.go
@@ -3,14 +3,17 @@ package networkpolicy
 import (
 	"context"
 	"fmt"
+	"runtime"
+	"strings"
 	"time"
+	"unique"
 
 	nfqueue "github.com/florianl/go-nfqueue"
 	"github.com/mdlayher/netlink"
-
 	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -173,6 +176,24 @@ func newController(client clientset.Interface,
 
 	// reduce memory usage only care about Labels and Status
 	trim := func(obj interface{}) (interface{}, error) {
+		if po, ok := obj.(*v1.Pod); ok {
+			ips := make([]v1.PodIP, 0, len(po.Status.PodIPs))
+			for _, i := range po.Status.PodIPs {
+				ips = append(ips, v1.PodIP{IP: intern(i.IP)})
+			}
+			return &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      intern(po.Name),
+					Namespace: intern(po.Namespace),
+					Labels:    internm(po.Labels),
+				},
+				Spec: v1.PodSpec{Hostname: intern(po.Spec.Hostname), NodeName: intern(po.Spec.NodeName)},
+				Status: v1.PodStatus{
+					Phase:  v1.PodPhase(intern(string(po.Status.Phase))),
+					PodIPs: ips,
+				},
+			}, nil
+		}
 		if accessor, err := meta.Accessor(obj); err == nil {
 			accessor.SetManagedFields(nil)
 		}
@@ -183,9 +204,16 @@ func newController(client clientset.Interface,
 		return nil, err
 	}
 
+	pods := 0
 	// process only local Pods that are affected by network policices
-	_, _ = podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: func(obj interface{}) {
+	_, _ = podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerDetailedFuncs{
+		AddFunc: func(obj interface{}, isInitialList bool) {
+			if isInitialList {
+				pods++
+				if pods%200 == 0 {
+					runtime.GC()
+				}
+			}
 			pod := obj.(*v1.Pod)
 			if pod.Spec.NodeName != c.config.NodeName {
 				return
@@ -774,4 +802,16 @@ func (c *Controller) cleanNFTablesRules() {
 	if err := c.nft.Run(context.TODO(), tx); err != nil {
 		klog.Infof("error deleting nftables rules %v", err)
 	}
+}
+
+func intern(s string) string {
+	return unique.Make(strings.Clone(s)).Value()
+}
+
+func internm(s map[string]string) map[string]string {
+	nm := make(map[string]string, len(s))
+	for k, v := range s {
+		nm[intern(k)] = intern(v)
+	}
+	return nm
 }
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/kube-network-policies
 
-go 1.22.0
+go 1.23.0
 
 require (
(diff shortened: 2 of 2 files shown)


pkg/networkpolicy/controller.go
```diff
@@ -173,6 +176,24 @@ func newController(client clientset.Interface,
 
 	// reduce memory usage only care about Labels and Status
 	trim := func(obj interface{}) (interface{}, error) {
+		if po, ok := obj.(*v1.Pod); ok {
+			ips := make([]v1.PodIP, 0, len(po.Status.PodIPs))
+			for _, i := range po.Status.PodIPs {
+				ips = append(ips, v1.PodIP{IP: intern(i.IP)})
+			}
+			return &v1.Pod{
```
aojea, line 184: is not replacing in place cheaper?

pkg/networkpolicy/controller.go
```diff
@@ -183,9 +204,16 @@ func newController(client clientset.Interface,
...
+	pods := 0
 	// process only local Pods that are affected by network policices
-	_, _ = podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc: func(obj interface{}) {
+	_, _ = podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerDetailedFuncs{
+		AddFunc: func(obj interface{}, isInitialList bool) {
+			if isInitialList {
+				pods++
+				if pods%200 == 0 {
+					runtime.GC()
```
aojea, line 214: this scares me

pkg/networkpolicy/controller.go
```diff
@@ -173,6 +176,24 @@ func newController(client clientset.Interface,
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      intern(po.Name),
+					Namespace: intern(po.Namespace),
+					Labels:    internm(po.Labels),
+				},
+				Spec: v1.PodSpec{Hostname: intern(po.Spec.Hostname), NodeName: intern(po.Spec.NodeName)},
+				Status: v1.PodStatus{
+					Phase:  v1.PodPhase(intern(string(po.Status.Phase))),
+					PodIPs: ips,
+				},
```
aojea, line 194: what I love to do is this server side, so you can trim the objects before sending them through the wire, there is a metadata informer only, but still need these values on Spec and Status

pkg/networkpolicy/controller.go
```diff
@@ -3,14 +3,17 @@ package networkpolicy
...
 	"fmt"
+	"runtime"
+	"strings"
 	"time"
+	"unique"
 
 	nfqueue "github.com/florianl/go-nfqueue"
 	"github.com/mdlayher/netlink"
-
 	v1 "k8s.io/api/core/v1"
```
danwinship, line 13: I prefer it with the blank line there.

(But we should adopt a consistent style. Currently:
  - `pkg/networkpolicy/controller.go` has a blank line between the github and k8s imports
  - `pkg/networkpolicy/metrics.go` puts them together
  - `cmd/main.go` is a mess

)

pkg/networkpolicy/controller.go
danwinship: ah, apparently `SetTransform` is the exception to the normal "you can't modify objects from the informer cache" rule

pkg/networkpolicy/controller.go
```diff
@@ -173,6 +176,24 @@ func newController(client clientset.Interface,
 
 	// reduce memory usage only care about Labels and Status
 	trim := func(obj interface{}) (interface{}, error) {
+		if po, ok := obj.(*v1.Pod); ok {
+			ips := make([]v1.PodIP, 0, len(po.Status.PodIPs))
+			for _, i := range po.Status.PodIPs {
+				ips = append(ips, v1.PodIP{IP: intern(i.IP)})
```
danwinship, line 182: Pod IPs don't get reused very quickly... This probably doesn't help _much_...
(Is there a performance tradeoff between "memory saved by interning" vs "extra allocations used in the process of interning"?)

pkg/networkpolicy/controller.go
```diff
@@ -173,6 +176,24 @@ func newController(client clientset.Interface,
...
+			}
+			return &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      intern(po.Name),
+					Namespace: intern(po.Namespace),
+					Labels:    internm(po.Labels),
+				},
+				Spec: v1.PodSpec{Hostname: intern(po.Spec.Hostname), NodeName: intern(po.Spec.NodeName)},
```
danwinship, line 190: the code doesn't seem to use `Hostname`

pkg/networkpolicy/controller.go
danwinship: > to do is this server side, so you can trim the objects before sending them through the wire

long term, the most efficient thing would be to have a central controller that processes the Pods and distributes processed data about them to the per-node kube-network-policy processes (in the same way kcm processes Pods into EndpointSlices so kube-proxy only needs to watch the latter). Using an API object (possibly even EndpointSlice) would be one way to do that, but using a grpc API (like xDS) would probably be better.

pkg/networkpolicy/controller.go
danwinship: it doesn't scare me but it seems kind of ugly and maybe would fit better in the the workqueue-processing code (eg `processNextItem`) rather than the informer itself?

pkg/networkpolicy/controller.go
aojea: it runs before goes to the informer cache IIRC

pkg/networkpolicy/controller.go
aojea: yeah, I started that path, and have a branch with that, but I didn't like moving to a two components models ... xDS is cool but in this case this is a cache synchronization problem, so current informers we'll be enough for efficient data transmission and will remove the additional dependencies

pkg/networkpolicy/controller.go
howardjohn: Yeah this is definitely likely to be dropped when I try to minimize the diff here. I did too much, mostly because I was misreading some pprofs before I added the explicitly `GC()` step. I doubt this adds any value.

pkg/networkpolicy/controller.go
howardjohn: The reason I didn't replace in place (and made sure to copy every item entirely) is due to how Go will sometimes not GC a full struct if you hold a reference to something within the struct. https://stackoverflow.com/a/55018900

However, I doubt this applies here, or if it is it can be avoided in a simpler manner -- I was measuring things wrong so took an overkill approach. ill clean it up

pkg/networkpolicy/controller.go
howardjohn: One other thing you could do is to have, on the client side, a custom Pod type that is only the subset. While the apiserver will send the full pod, it will only end up as bytes and not deserialized into a Pod object. I did similar a while back with decent results (but it wasn't with Kubernetes; its likely a pain to do with client-go).

Then you probably don't need the GC thing either. 

Obviously server side is best though

pkg/networkpolicy/controller.go
paulgmiller: Maybe GOMEMLIMIT  would be a more conventional approach?

PR #95, opened 2024-10-07
Workaround dns kernel bug

aojea:
Add a workaround to the netfilter conntrack bug https://bugzilla.netfilter.org/show_bug.cgi?id=1766 fixed by https://github.com/torvalds/linux/commit/8af79d3edb5fd2dce35ea0a71595b6d4f9962350.

Since the race problem is caused by having two packets DNATed with the same tuple at the same time by different CPUs, it impact specially DNS resolvers that sends A and AAAA request in parallel (it seems is default glibc behavior), and this is very visible to users that notice DNS latency.

Instead of processing all packets on the POSROUTING hook, special case DNS and process them in PREROUTING after DNAT happens.

Services implementations that don't rely on conntrack can disable this behavior by setting the flag to false.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -55,7 +55,9 @@ func main() {
 	klog.InitFlags(nil)
 	flag.Parse()
 
-	klog.Infof("flags: %v", flag.Args())
+	flag.VisitAll(func(flag *flag.Flag) {
+		klog.Infof("FLAG: --%s=%q", flag.Name, flag.Value)
+	})
 
 	if _, _, err := net.SplitHostPort(metricsBindAddress); err != nil {
 		klog.Fatalf("error parsing metrics bind address %s : %v", metricsBindAddress, err)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -34,6 +34,7 @@ var (
 	queueID                    int
 	metricsBindAddress         string
 	hostnameOverride           string
+	netfilterBug1766Fix        bool
 )
 
 func init() {
@@ -43,6 +44,7 @@ func init() {
 	flag.IntVar(&queueID, "nfqueue-id", 100, "Number of the nfqueue used")
 	flag.StringVar(&metricsBindAddress, "metrics-bind-address", ":9080", "The IP address and port for the metrics server to serve on")
 	flag.StringVar(&hostnameOverride, "hostname-override", "", "If non-empty, will be used as the name of the Node that kube-network-policies is running on. If unset, the node name is assumed to be the same as the node's hostname.")
+	flag.BoolVar(&netfilterBug1766Fix, "netfilter-bug-1766-fix", true, "If set, process DNS packets on the PREROUTING hooks to avoid the race condition on the conntrack subsystem, not needed for kernels 6.12+ (see https://bugzilla.netfilter.org/show_bug.cgi?id=1766)")
 
 	flag.Usage = func() {
 		fmt.Fprint(os.Stderr, "Usage: kube-network-policies [options]\n\n")
@@ -74,6 +76,7 @@ func main() {
 		FailOpen:                   failOpen,
 		QueueID:                    queueID,
 		NodeName:                   nodeName,
+		NetfilterBug1766Fix:        netfilterBug1766Fix,
 	}
 	// creates the in-cluster config
 	config, err := rest.InClusterConfig()
--- a/pkg/networkpolicy/controller.go
+++ b/pkg/networkpolicy/controller.go
@@ -681,83 +681,82 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 	// IPVS packets follow a different path in netfilter, so we process
 	// everything in the POSTROUTING hook before SNAT happens.
 	// Ref: https://github.com/kubernetes-sigs/kube-network-policies/issues/46
-	for _, hook := range []knftables.BaseChainHook{knftables.PostroutingHook} {
-		chainName := string(hook)
-		tx.Add(&knftables.Chain{
-			Name:     chainName,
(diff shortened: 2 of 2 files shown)


aojea:
see if @danwinship wants to take a look, otherwise I will merge end of the day, so we can have it in kind

cmd/main.go
```diff
@@ -43,6 +44,7 @@ func init() {
 	flag.IntVar(&queueID, "nfqueue-id", 100, "Number of the nfqueue used")
 	flag.StringVar(&metricsBindAddress, "metrics-bind-address", ":9080", "The IP address and port for the metrics server to serve on")
 	flag.StringVar(&hostnameOverride, "hostname-override", "", "If non-empty, will be used as the name of the Node that kube-network-policies is running on. If unset, the node name is assumed to be the same as the node's hostname.")
+	flag.BoolVar(&netfilterBug1766Fix, "netfilter-bug-1766-fix", true, "If set, process DNS packets on the PREROUTING hooks to avoid the race on the conntrack subsystem, not needed for kernels 6.12+ (see https://bugzilla.netfilter.org/show_bug.cgi?id=1766)")
```
jglick, line 47:
```suggestion
	flag.BoolVar(&netfilterBug1766Fix, "netfilter-bug-1766-fix", true, "If set, process DNS packets on the PREROUTING hooks to avoid the race condition on the conntrack subsystem, not needed for kernels 6.12+ (see https://bugzilla.netfilter.org/show_bug.cgi?id=1766)")
```
?

pkg/networkpolicy/controller.go
```diff
@@ -681,83 +682,97 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
+		Name: chainName,
+	})
+
+	// DNS is processed by addDNSRacersWorkaroundRules()
+	// TODO: remove once kernel fix is on most distros
+	if c.config.NetfilterBug1766Fix {
+		tx.Add(&knftables.Rule{
+			Chain:   chainName,
+			Rule:    "udp dport 53 accept",
+			Comment: ptr.To("process DNS traffic on PREROUTING hook with network policy enforcement to avoid netfiler race bug"),
```
jglick, line 703:
```suggestion
			Comment: ptr.To("process DNS traffic on PREROUTING hook with network policy enforcement to avoid netfilter race bug"),
```
or

```suggestion
			Comment: ptr.To("process DNS traffic on PREROUTING hook with network policy enforcement to avoid netfilter race condition bug"),
```

pkg/networkpolicy/controller.go
```diff
@@ -767,6 +782,74 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 	return nil
 }
 
+// To avoid a kernel bug caused by UDP DNS request racing with conntrack
+// process the DNS packets only on the PREROUTING hook after DNAT happens
+// so we can see the resolved destination IPs, typically the ones of the Pods
+// that are used for the Kubernetes DNS Service.
+// xref: https://github.com/kubernetes-sigs/kube-network-policies/issues/12
+// This can be removed once all kernels contain the fix in
+// https://github.com/torvalds/linux/commit/8af79d3edb5fd2dce35ea0a71595b6d4f9962350
```
jglick, line 791: (https://github.com/torvalds/linux/commit/8af79d3edb5fd2dce35ea0a71595b6d4f9962350 for reference)

pkg/networkpolicy/controller.go
```diff
@@ -767,6 +782,74 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
+			),
+			Comment: ptr.To("process IPv6 traffic destined to a DNS serverwith network policy enforcement"),
+		})
+
+		tx.Add(&knftables.Rule{
+			Chain: chainName,
+			Rule: knftables.Concat(
+				"ip6", "daddr", "@", podV6IPsSet, "udp dport 53", action,
+			),
+			Comment: ptr.To("process IPv6 traffic destined to a DNS serverwith network policy enforcement"),
```
jglick, line 841:
```suggestion
			Comment: ptr.To("process IPv6 traffic destined to a DNS server with network policy enforcement"),
```

jglick (commented):
Just noting a couple typos in comments. Thanks again for all your work on this!

pkg/networkpolicy/controller.go
```diff
@@ -123,7 +123,7 @@ func newController(client clientset.Interface,
 		nft:    nft,
 		queue: workqueue.NewTypedRateLimitingQueueWithConfig(
 			workqueue.DefaultTypedControllerRateLimiter[string](),
-			workqueue.TypedRateLimitingQueueConfig[string]{Name: "controllerName"},
+			workqueue.TypedRateLimitingQueueConfig[string]{Name: controllerName},
```
danwinship, line 127: 1. lol
2. commit message ("bump dependencies") does not match the commit body?

danwinship (commented):
where's the regression test? :slightly_smiling_face:

pkg/networkpolicy/controller.go
aojea: 1 is something we have in all kubernetes components :) , the os.Args does not give the passed flags. it is very useful for debugging

2. now I understand why I have a merge conflict, it seems I have an old version, and once I rebased over the main  I got the dependency update :)

aojea:
> where's the regression test? 🙂

do you mean in kubernetes/kubernetes? this showed up in this e2e https://github.com/kubernetes-sigs/kube-network-policies/issues/12
but it is a bug in the kernel, and its a race, there is a kselftest, but I don't see what type of tests we want to add here

PR #101, opened 2024-10-22
Update install-anp.yaml

aojea:
don't use the default queue number to avoid conflict with kindnet

### DIFF
--- a/install-anp.yaml
+++ b/install-anp.yaml
@@ -80,7 +80,7 @@ spec:
       - name: kube-network-policies
         image: registry.k8s.io/networking/kube-network-policies:v0.6.0
         command: ["/bin/netpol"]
-        args: ["-v=2", "-admin-network-policy=true", "-baseline-admin-network-policy=true","--hostname-override=$(MY_NODE_NAME)"]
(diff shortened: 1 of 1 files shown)


install-anp.yaml
```diff
@@ -80,7 +80,7 @@ spec:
       - name: kube-network-policies
         image: registry.k8s.io/networking/kube-network-policies:v0.6.0
         command: ["/bin/netpol"]
-        args: ["-v=2", "-admin-network-policy=true", "-baseline-admin-network-policy=true","--hostname-override=$(MY_NODE_NAME)"]
+        args: ["-v=2", "-admin-network-policy=true", "-baseline-admin-network-policy=true","--hostname-override=$(MY_NODE_NAME)","--nfqueue-id=99"]
```
tssurya, line 83: lgtm, since default I guess is 0 what kube-net-pol uses will conflict with that kindnet uses so safer to use a custom value..

aojea: default is 100

https://github.com/kubernetes-sigs/kube-network-policies/blob/d3eeb5525d5a47b8e4601ecc3476b120b0156992/cmd/main.go#L44

tssurya: so kube-net-pol in KIND uses 100 and hence the conflict... when we install it separately.. I see

PR #104, opened 2024-10-22
Update npa.yml

aojea:


tssurya:
at least its failing here consistently so we know this is the problem :) thanks @aojea !

aojea:
> at least its failing here consistently so we know this is the problem :) thanks @aojea !

let's rebase and retry once the others merged

PR #105, opened 2024-10-22
Update e2e jobs

aojea:


.github/workflows/npa.yml
```diff
@@ -13,8 +13,8 @@ on:
 
 env:
   GO_VERSION: "1.22.0"
-  K8S_VERSION: "v1.30.0"
-  KIND_VERSION: "v0.23.0"
+  K8S_VERSION: "v1.30.4"
+  KIND_VERSION: "v0.24.0"
```
aojea, line 17: ok, so is the kind version that is the first one with network policies embedded ... it seems there is some bad interactions

cmd/main.go
```diff
@@ -77,6 +77,7 @@ func main() {
 		QueueID:                    queueID,
 		NodeName:                   nodeName,
 		NetfilterBug1766Fix:        netfilterBug1766Fix,
+		NFTableName:                "kube-network-policies-different",
```
aojea, line 80: @tssurya if my theory is correct, the problem is that kindnet and kube-network-policies are using the same table and stomping into each other

cc: @danwinship

aojea:
jobs are passing now, let me split the commit with the fix, so we can merge and revendor in kindnet, before we do a new kind release

danwinship:
/lgtm
/hold
(I wasn't sure if your comment above meant to not merge this yet. Feel free to un-hold.)

aojea:
> /lgtm /hold (I wasn't sure if your comment above meant to not merge this yet. Feel free to un-hold.)

/hold cancel

now is good, @tssurya you should mimic this change to use the latest kind version and node image

Thanks

tssurya:
> > /lgtm /hold (I wasn't sure if your comment above meant to not merge this yet. Feel free to un-hold.)
> 
> /hold cancel
> 
> now is good, @tssurya you should mimic this change to use the latest kind version and node image
> 
> Thanks

ack on it, I will update my other PR in network-policy-api repo

PR #106, opened 2024-10-23
Allow to configure the nftables name and implement defaulting in the config data structure

aojea:


pkg/networkpolicy/controller.go
```diff
@@ -67,6 +68,24 @@ type Config struct {
...
+	NFTableName                string // if other projects use this controllers they need to be able to use their own table name
+}
+
+func (c *Config) Defaults() error {
+	var err error
+	if c.QueueID == 0 {
+		c.QueueID = 100
+	}
+	if c.NodeName == "" {
+		c.NodeName, err = os.Hostname()
```
tssurya, line 80: sorry I am not familiar with this.. what is the NodeName here used for?

pkg/networkpolicy/controller.go
```diff
@@ -67,6 +68,24 @@ type Config struct {
...
+		if err != nil {
+			return err
+		}
+	}
+	if c.NFTableName == "" {
```
tssurya, line 85: I was about to ask for "do you have docs for config options" but then I saw the nice:
```
	flag.IntVar(&queueID, "nfqueue-id", 100, "Number of the nfqueue used")
```
flags one can pass, should we add the new tableName to that? or is there a difference between internal and external configs?

pkg/networkpolicy/controller_test.go
```diff
@@ -125,3 +127,70 @@ func newTestController() *networkpolicyController {
...
+func TestConfig_Defaults(t *testing.T) {
+	tests := []struct {
+		name     string
+		config   Config
+		expected Config
+	}{
+		{
+			name: "empty",
+			config: Config{
+				NodeName: "testnode", // nodename defaults to os.Hostname so we ignore for tests
```
tssurya, line 140: hmm the comment is confusing probably because I miss context, if "testnode" is an explicit override you are providing, then we are not ignoring for tests right?

tssurya (commented):
overall lgtm, a few inline questions..

pkg/networkpolicy/controller.go
aojea: The network policy implementation , for being more efficient, only apply policis to the Pods that are local (same node)

pkg/networkpolicy/controller.go
aojea: yeah, I don't want to expose the table name to the users of kube-network-policies , though I want to expose it to the consumer of the controller, in this case kindnet

pkg/networkpolicy/controller_test.go
aojea: if is set, then it will not be defaulted :) .. the test is about testing the Defaults are applied

pkg/networkpolicy/controller.go
tssurya: a I see :) makes sense thanks!

pkg/networkpolicy/controller_test.go
tssurya: ack so testing os.Hostname is ignored is what you mean given we'd need mocks etc so you do the overrides.. gotcha!

tssurya (approved):
/lgtm

pkg/networkpolicy/controller_test.go
aojea: yeah, not great comment and good review, but I think we can live with it

Thanks

Issue #98, opened 2024-10-11
Provide helm chart


aojea:
I had this feedback that it will be useful to provide a helm chart for users

mneverov:
@aojea Should CRDs be included in the chart?

aojea:
not familiar with the helm ecosystem and its practices, but the CRDs , if included should not be installed by default

mneverov:
implemented in #107, can be closed

aojea:
Thank you very much

PR #107, opened 2024-10-25
Add helm chart.

mneverov:
Addresses #98 

To test run `helm install kube-network-policies -n kube-system charts/kube-network-policies`.
`admin-network-policy` and `baseline-admin-network-policy` are [enabled by default](https://github.com/kubernetes-sigs/kube-network-policies/compare/main...mneverov:kube-network-policies:helm?expand=1#diff-80cda0676b5576ede7ef733618470247b0b084a60de2300c0ec3b7e6fd1b63bbR9) (values file `baselineAdminNetworkPolicy: true`).

### DIFF
--- a/charts/kube-network-policies/templates/daemonset.yaml
+++ b/charts/kube-network-policies/templates/daemonset.yaml
@@ -1,0 +1,75 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ template "kube-network-policies.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "kube-network-policies.labels" . | nindent 4 }}
+    {{- with .Values.daemonset.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.daemonset.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "kube-network-policies.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "kube-network-policies.labels" . | nindent 8 }}
+        {{- with .Values.daemonset.labels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      hostNetwork: true
+      dnsPolicy: ClusterFirst
+      {{- with .Values.daemonset.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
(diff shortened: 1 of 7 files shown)


aojea:
awesome, do you mind add a sentence in https://github.com/kubernetes-sigs/kube-network-policies/blob/4320aa92e3f00b4951d11bc88081cfb2ff9f3f72/README.md?plain=1#L7 to mention this new install method?

charts/kube-network-policies/templates/daemonset.yaml
```diff
@@ -0,0 +1,75 @@
...
+      {{- end }}
+      serviceAccountName: {{ include "kube-network-policies.serviceAccountName" . }}
+      containers:
+        - name: kube-network-policies
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          args:
+            - /bin/netpol
+            - --hostname-override=$(MY_NODE_NAME)
+            - --v=2
+            {{- if .Values.baselineAdminNetworkPolicy }}
```
aojea, line 45:
```suggestion
            {{- if .Values.adminNetworkPolicy }}
```

I learned this recently, so there is Admin Network Policy, that has two CRD adminNetworkPolicy and baseAdminNetworkPolicy , so, maybe better to name `Values.adminNetworkPolicy`     instead of `Values.baselineAdminNetworkPolicy`   ??

aojea:
@mneverov naive question, the admin network policies require to install some CRDs, does helm practices recommend to embed those CRDs?

https://github.com/kubernetes-sigs/kube-network-policies/blob/4320aa92e3f00b4951d11bc88081cfb2ff9f3f72/.github/workflows/npa.yml#L110-L111

charts/kube-network-policies/templates/daemonset.yaml
mneverov: You're right.
JFYI, we specify both flags in [install-anp](https://github.com/kubernetes-sigs/kube-network-policies/blob/main/install-anp.yaml#L83C24-L83C88). I could not come up with a generic parameter name that describes both AdminNetworkPolicy and BaselineAdminNetworkPolicy.
Renamed as suggested.

mneverov:
> @mneverov naive question, the admin network policies require to install some CRDs, does helm practices recommend to embed those CRDs?

For a user convenience CRDs are often embedded in the chart especially when they are defined in the same repo. CRDs can also be installed as a [dependency](https://helm.sh/docs/helm/helm_dependency/), but in this case they should be also provided as a helm chart. I don't see helm charts for CRDs in [network-policy-api](https://github.com/kubernetes-sigs/network-policy-api), so users still need to install them manually.

aojea:
This is awesome, thank you very much

/LGTM
/approve

PR #115, opened 2024-11-14
Give metrics and logging a verdict dimension

paulgmiller:
Playing around with testing and wanted have a little more info on packets per verdict

### DIFF
--- a/pkg/networkpolicy/controller.go
+++ b/pkg/networkpolicy/controller.go
@@ -460,8 +460,9 @@ func (c *Controller) Run(ctx context.Context) error {
 			processingTime := float64(time.Since(startTime).Microseconds())
 			packetProcessingHist.WithLabelValues(string(packet.proto), string(packet.family)).Observe(processingTime)
 			packetProcessingSum.Observe(processingTime)
-			packetCounterVec.WithLabelValues(string(packet.proto), string(packet.family)).Inc()
-			logger.V(2).Info("Finished syncing packet", "id", *a.PacketID, "duration", time.Since(startTime), "accepted", verdict == nfqueue.NfAccept)
+			verdictStr := verdictString(verdict)
+			packetCounterVec.WithLabelValues(string(packet.proto), string(packet.family), verdictStr).Inc()
+			logger.V(2).Info("Finished syncing packet", "id", *a.PacketID, "duration", time.Since(startTime), "verdict", verdictStr)
 		}()
 
 		if c.evaluatePacket(ctx, packet) {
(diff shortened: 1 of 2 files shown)


aojea:
/ok-to-test
Thanks

pkg/networkpolicy/metrics.go
```diff
@@ -33,7 +33,7 @@ var (
 	packetCounterVec = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "packet_count",
 		Help: "Number of packets",
-	}, []string{"protocol", "family"})
+	}, []string{"protocol", "family", "verdict"})
```
paulgmiller, line 36: If we tossed the tossed the whole packet 5 part tuple in here,  we  could use a cli tool to generate network policies for the last amount of time all your daemonsets were up. But that seem like maybe an optional metric to leave on so leaving for another PR.

pkg/networkpolicy/controller.go
```diff
@@ -458,8 +458,9 @@ func (c *Controller) Run(ctx context.Context) error {
 			processingTime := float64(time.Since(startTime).Microseconds())
 			packetProcessingHist.WithLabelValues(string(packet.proto), string(packet.family)).Observe(processingTime)
 			packetProcessingSum.Observe(processingTime)
-			packetCounterVec.WithLabelValues(string(packet.proto), string(packet.family)).Inc()
-			klog.V(2).Infof("Finished syncing packet %d took: %v accepted: %v", *a.PacketID, time.Since(startTime), verdict == nfqueue.NfAccept)
+			verdictStr := verdictString(verdict)
+			packetCounterVec.WithLabelValues(string(packet.proto), string(packet.family), verdictStr).Inc()
+			klog.V(2).Infof("Finished syncing packet %d took: %v verdict: %v", *a.PacketID, time.Since(startTime), verdictStr)
```
paulgmiller, line 463: This conflicts with https://github.com/kubernetes-sigs/kube-network-policies/pull/50 so will wait on @uablrek if that's inbound

uablrek: I re-based #50

aojea: should we close this now that #50 merged?

paulgmiller:
This seems to be working as intended when I toss it on a cluster

# TYPE packet_count counter
packet_count{family="IPv4",protocol="TCP",verdict="accept"} 627
packet_count{family="IPv4",protocol="TCP",verdict="drop"} 13

pkg/networkpolicy/controller.go
```diff
@@ -493,6 +494,19 @@ func (c *Controller) Run(ctx context.Context) error {
...
+// verifctString coverts nfqueue int vericts to strings for metrics/logging
+// it does not cover all of them because we should only use a subset.
+func verdictString(verdict int) string {
+	switch verdict {
+	case nfqueue.NfDrop:
+		return "drop"
+	case nfqueue.NfAccept:
+		return "accept"
+	default:
+		return "unknonw"
```
aojea, line 506:
```suggestion
		return "unknown"
```

aojea:
one typo to fix and ready to merge

pkg/networkpolicy/metrics.go
paulgmiller: Yep exactly and it doesn't even catch the labels so would almost certainly have to come through logs to build something like this. 

Verdict is still helpful though as people can alert on continuos blocking as its either a misconfiguration or an attack.

aojea: agree

aojea:
/lgtm
/approve

Thanks

Issue #99, opened 2024-10-11
Improve documentation


aojea:
There is a README but is outdated

It will be also nice to have some kind of static web as https://network-policy-api.sigs.k8s.io/api-overview/

kenudeh:
Hello there! By static web, you mean a proper documentation site, right?

kenudeh:
Just reviewed the `/docs/testing` folder, there doesn't seem to be any documentation at all for the project at the moment. Is this correct?

PR #119, opened 2024-11-18
[Docs] comment on udp and effected pods in readme

paulgmiller:
Wasn't clear that udp didn't get send every packet to user space so after I confirmed that established also ignored tcp flows I thought I'd update the readme. 

Also the readme made it seem like ignoring non selected pods was a future optimization but it seems already one so I updated that paragraph too. 

Somewhat related to #99


aojea:
/ok-to-test
/lgtm
/approve

Thanks

Issue #121, opened 2024-11-20
Further Separate Packet Interception From Verdict Rendering


paulgmiller:
One of the things that makes me happiest about this project is that evaluatePacket pretty cleanly separates network policy logic from underlying implementation. Its essentially give me a packet and I will use K8s objects to decide its fate. 

The rest of the controller however is pretty coupled to nfqueue. I am wondering if we might benefit from having the controller be completely agnostic to how packet interception happens. A concrete suggestion to do this would be to have the controller take an interface for interception that does two things. 
1. Calls evaluatePacket on each packet it decides to intercept
2. Takes advice on what local ips it should intercept  (sync below) 
Very open to refining this interface. 
```
type interceptor interface {
	Run(context.Context, func(Packet) int) error
	Sync(ctx context.Context, podV4IPs, podV6IPs sets.Set[string]) error
}
```

This would allow 
1) controller to reduce its scope and nfqueue logic to live in its own package.
2) Other interceptors to be tried out without bothering nfqueue or the core verdict logic. This might be just a future breaking api change in nftables, a windows implementation (maybe https://github.com/tailscale/wf), or another tech completely like ebpf. 

Tried putting something together that would pass unittests and it seems doable so wanted start an issue to maybe spark a conversation here or in sig apps slack.

aojea:
sounds reasonable

paulgmiller:
Slowed down by massive seattle power outage but alrady realized the interface may need a Stop() or has to return a function to defer cleanup. Could also make main defer the cleanup 

I'm usually a fan of you don't know what an interface should look like till you have 3 users so I might go play with the windows packet filter just to seee if it fits this interface then come back with adjustments. Refactor is already plenty large though so I would not actually include any implemntations in it (and may try and remove some unceesary changes I made)

danwinship:
> If we can expose, let's say a grpc service, that receives a tuple and emits a verdict, all the agents just need to send the request to this new service and will be very dumb and with very low memory footprint, and all the complexity will move to this intermediate agents

But that would also be much slower; establishment of new connections now has to wait for a separate gRPC request to succeed first. And if the individual nodes know _nothing_ about the specific NetworkPolicy rules, then that means they need to intercept _every_ pod-to-X or X-to-pod connection and send it to the gRPC server, rather than being able to optimize their ruleset to only intercept traffic that might potentially get blocked.

aojea:
[...]

PR #122, opened 2024-11-20
Seperate packet interception into seperate interface/package to get it out of controller. 

paulgmiller:
Described alot more of intent in #121 

Very much a "lets see if this makes sense" rather than  "please review for merging" PR.


pkg/networkpolicy/controller.go
```diff
@@ -347,9 +341,18 @@ type Controller struct {
...
 	getPodAssignedToIP func(podIP string) *v1.Pod
 
-	nft     knftables.Interface // install the necessary nftables rules
-	nfq     *nfqueue.Nfqueue
-	flushed bool
+	interceptor interceptor
+}
+
+// Verdicts
+const (
```
paulgmiller, line 348: should this be a type rather than an int?

aojea:
I see you found a lot of cleanups, can you please split those commits and send in a different PR ... the Interface looks pretty neat, seems a good idea

aojea:
/reopen

@paulgmiller I really like to progress in this direction, as this can solve the use case for network policies across clusters and clouds with existing NetworkPolicies APIs, and no need to develop new APIs. Let me elaborate on it:

Requirements:
- clusters IPAM does not overlap, each Pod is guarantee to have unique IPs across cluster
- existing APIs has the same meaning across cluster, i.e: namespace foo in cluster A from the network policy perspective has  the same meaning as in cluster B, so a network policy that selects that namespace applies to all clusters
- clusters are differentiated through a special label, `network-policy.x-k8s.io/cluster: clusterA`, how to identify clusters is left for discussion, but network policies can use this label to apply 
- clusters are able to talk between each other
-
Strawman approach:
1. we have a global database with a well defined schema that contains all clusters Pod and Namespaces metadata information, meanwhile the schema is common this allows to use different plugins and use different global db backends, specially useful for cloud provider that can use a plugin for spanner, dynamodb, cockroachdb, cosmosdb, ...
2. we create some new special agents on the nodes to run as a deployment, so we can scale out to handle scalability with a double functionality:
  2.1. Sync existing cluster state with the database
  2.2 Expose a Service that emits a verdict for the kube-network-policy agent, it receives a packet tuple and checks against the local information if it should be dropped or accepted

This interface in the PR is the key for implementing 2.2, if we can expose, let's say a grpc service, that receives a tuple and emits a verdict,  all the agents just need to send the request to this new service and will be very dumb and with very low memory footprint, and all the complexity will move to this intermediate agents

cc: @thockin @danwinship @wojtek-t

PR #156, opened 2025-03-18
[RFC] pkg/networkpolicy: Don't require root for TestNetworkPolicies_SyncRules

rata:
Here is a PoC of using userns to not require root privileges to run the tests. I was talking about this with  @aojea 

The first commit is quite a small diff to show what is really needed. Please see that commit separately. The other commit detects if unprivileged userns is enabled and fallbacks to run as root otherwise. But the first commit is "the bare minimum".

~If there is interest to use this, this should be turned probably into a more generic helper and then adopted on more tests. But this is just a PoC :)~
EDIT: Oh, it seems this the only test that needs root? So maybe this could be adopted with some, but not so much more massaging :)

Here is the commit msg for the first commit:

---
We can run the unit test without requiring root privileges if we use user namespaces.

Some distros restrict unprivileged users to create a user namespace, so this will need tuning for those cases. One known case is ubuntu 24.04, we need to run this to enable it:

	sysctl -w kernel.apparmor_restrict_unprivileged_userns=0

### DIFF
--- a/pkg/networkpolicy/controller_test.go
+++ b/pkg/networkpolicy/controller_test.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"testing"
 
@@ -19,6 +20,11 @@ import (
 	"k8s.io/client-go/tools/cache"
 	npaclientfake "sigs.k8s.io/network-policy-api/pkg/client/clientset/versioned/fake"
 	npainformers "sigs.k8s.io/network-policy-api/pkg/client/informers/externalversions"
+)
+
+var (
+	usernsEnabled bool
+	checkUserns   sync.Once
 )
 
 func makeNamespace(name string) *v1.Namespace {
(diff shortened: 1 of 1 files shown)


rata:
Heh, I'm member of the k8s org, but not k8s-sigs, so I need the ok to test

aojea:
/ok-to-test
/cc @BenTheElder @adrianmoisey 

Thanks, let's see

rata:
@aojea I GUESS the CI is probably running as root today already, so we will need to change that for this to take effect. Maintainers can push to this branch on my fork, so feel free to push changes to not run CI as root (and if using ubuntu 24.04, remember the sysctl)

aojea:
This is awesome

```
go test -v ./pkg/networkpolicy/ -run TestNetworkPolicies_SyncRules
=== RUN   TestNetworkPolicies_SyncRules
    controller_test.go:242: === RUN   TestNetworkPolicies_SyncRules
        === RUN   TestNetworkPolicies_SyncRules/subprocess
        === RUN   TestNetworkPolicies_SyncRules/subprocess/default
        I0318 17:39:32.369955 3742077 controller.go:669] "Syncing nftables rules"
        === RUN   TestNetworkPolicies_SyncRules/subprocess/admin_network_policy
        I0318 17:39:32.434293 3742077 controller.go:669] "Syncing nftables rules"
        === RUN   TestNetworkPolicies_SyncRules/subprocess/bug_disabled
        I0318 17:39:32.494264 3742077 controller.go:669] "Syncing nftables rules"
        --- PASS: TestNetworkPolicies_SyncRules (0.20s)
            --- PASS: TestNetworkPolicies_SyncRules/subprocess (0.20s)
                --- PASS: TestNetworkPolicies_SyncRules/subprocess/default (0.06s)
                --- PASS: TestNetworkPolicies_SyncRules/subprocess/admin_network_policy (0.06s)
                --- PASS: TestNetworkPolicies_SyncRules/subprocess/bug_disabled (0.08s)
        PASS
--- PASS: TestNetworkPolicies_SyncRules (0.22s)
PASS
ok      sigs.k8s.io/kube-network-policies/pkg/networkpolicy     0.287s
```

/lgtm
/approve

BenTheElder (commented):
/hold

BenTheElder:
I think we first need to define something similar to integration tests or e2e tests so this doesn't necessarily run in `make test`, because it's reaching out and mutating global state on the machine.

Also, regardless, these need to be constrained to linux only, which they do not appear to be currently.

BenTheElder:
> @aojea I GUESS the CI is probably running as root today already, so we will need to change that for this to take effect. Maintainers can push to this branch on my fork, so feel free to push changes to not run CI as root (and if using ubuntu 24.04, remember the sysctl)

No, unit tests do not run as root.

BenTheElder:
/hold-cancel

☕ 
Apologies, antonio pinged this in slack where we were discussing previously testing k/k.

We do not run unit tests as root *in Kubernetes* and I would expect more discussion there before moving forward. I will not for this repo 😅

BenTheElder:
https://prow.k8s.io/command-help#hold

/hold cancel
(or `/unhold`, `/remove-hold` but not `/hold-cancel`)

Issue #32, opened 2024-06-04
Implement DomainNames Egress Peer in ANP API


tssurya:
1.  We would love kube-network-policies to implement https://network-policy-api.sigs.k8s.io/npeps/npep-133-fqdn-egress-selector/
2. This API NPEP merged last week AND we will soon have the API merged and we don't yet have any implementations
3. So its also hard to write conformance tests without a testing driver backend

API Details:

- https://network-policy-api.sigs.k8s.io/npeps/npep-133-fqdn-egress-selector/#api -> It is an egress only peer that is a list of domain names
- https://network-policy-api.sigs.k8s.io/npeps/npep-133-fqdn-egress-selector/#goals -> only supports "allow action" since implementing deny with domain names cannot be enforced correctly

So implementing it here will help us make it part of our upstream CI tests.

tssurya:
@rahulkjoshi : The API PR will follow soon right?

vaskozl:
One thing I'm wondering about the NPEP:

If a policy is applied after a lookup is already performed and cached by the client, is the implementor required to allow the traffic based on the past query?

aojea:
> * This API NPEP merged last week AND we will soon have the API merged and we don't yet have any implementations

@tssurya please ping me once the API PR is ready for review so we can give feedback from the implementation during the review

tssurya:
API PR has merged, @aojea and I spoke at KubeCon and we will implement it.

aojea:
@tssurya what tag should I vendor https://github.com/kubernetes-sigs/network-policy-api/tags?

PR #157, opened 2025-04-08
Admin Network Policies: NPEP-133: Domain Names Selector for Egress Traffic

aojea:
Implements https://network-policy-api.sigs.k8s.io/npeps/npep-133-fqdn-egress-selector/

We just need to build a DNS cache wtih the responses and iterate over the peer rules to match on the resolved IPs

There are some tricky parts here, main one is on how to obtain the DNS information, we can use netfilter for that, since it offers to APIs: nflog for passive monitoring, nfqueue for active.

First try with nflog was succesful and less risky, but it was racy, it turns out nflog is slow and the dns information was obtained AFTER the client sends the request, making the effort useless. Nfqueue does not have that problem since allows to "hold" the packet meanwhile we process it, resolving the race problem. On the angle of performance vs security , I highly doubt that people will prefer to have the race and random behaviors vs the penalty that passing the packet to userspace that is in the order of O(us). The nfqueue is configured to fail open, so any issues with the daemon will not impact the cluster.

On the implementation side, for the DNS wildcard I opted by a radix tree for prefix matching, since seems to be the most performant solution, just bear in mind you need to reverse the domains since the wild card is for the suffix (*.kubernetes.io matches blog.kubernetes.io)

Another tricky part is the dependency on the Pod.Status.PodIPs, since those are updated after the pod is created by the kubelet in the apiserver, the network policy agent needs to wait for that roundtrip, using NRI the agent can get the PodIPs at the Pod creation removing also this race, so this fixes this problem https://github.com/kubernetes/kubernetes/issues/85966

Fixes: https://github.com/kubernetes-sigs/kube-network-policies/issues/32


pkg/networkpolicy/dnsagent.go
```diff
@@ -0,0 +1,252 @@
...
+	  [ log group 100 snaplen 0 qthreshold 0 ]
+	*/
+	nft.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: chain,
+		Exprs: []expr.Any{
+			&expr.Meta{Key: expr.MetaKeyL4PROTO, SourceRegister: false, Register: 0x1},
+			&expr.Cmp{Op: 0x0, Register: 0x1, Data: []byte{syscall.IPPROTO_UDP}},
+			&expr.Payload{DestRegister: 0x1, Base: expr.PayloadBaseTransportHeader, Offset: 0, Len: 2},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 0x1, Data: binaryutil.BigEndian.PutUint16(53)},
```
danwinship, line 141: Given the chain definition, this rule runs on the post-DNAT packet, right? You can't assume port 53 then. (OpenShift uses `targetPort: 5353` on the CoreDNS pods so they don't need `CAP_NET_BIND_SERVICE`.)

The equivalent of `ct original proto-dst` should work.

aojea: it is the postrouting on the return packet, so is the answer the pod will receive ... my thinking is

pod a --- > dns port 53

we capture just before it gets delivered to the pod, so it has to have src port 53 at that time

pkg/networkpolicy/networkpolicyapi.go
```diff
@@ -60,6 +60,12 @@ func (c *Controller) evaluateAdminEgress(adminNetworkPolices []*npav1alpha1.Admi
 						return rule.Action
 					}
 				}
+
+				for _, domain := range to.DomainNames {
+					if c.dnsCache.containsIP(string(domain), ip) {
+						return rule.Action
+					}
+				}
```
aojea, line 68: @danwinship @tssurya  is not clear to me from the API how this should be implemented, as it says that only allow rules are permitted, does it mean this is correct?

aojea:
@danwinship @tssurya some help here, do you mind creating an test case for the unit tests in https://github.com/kubernetes-sigs/kube-network-policies/blob/main/pkg/networkpolicy/networkpolicyapi_test.go and paste the patch so I can incorporate?

I need to focus on the dns packet capture logic

pkg/networkpolicy/networkpolicyapi.go
aojea: ok, then this is ok, if the validation gets relaxed this will do the right thing

aojea:
all green 🤟

PR #164, opened 2025-04-16
Fix graceful shutdown: handle SIGTERM properly to cleanup resources

gsakun:
## What this PR does

This PR fixes the graceful shutdown behavior of the application running in Kubernetes.

Previously, when the Pod was deleted, the application did not properly handle the termination signal", resulting in:
- No shutdown logs being printed.
- Residual nftables rules remaining on the host.

## What has been changed

- Introduced `signal.NotifyContext` to properly capture `SIGTERM` and `SIGINT`.
- Ensured cleanup logic is executed upon receiving shutdown signal, including nftables teardown.
- Added logs to indicate shutdown progress and completion.

## Testing

Tested by deploying to Kubernetes and deleting the Pod. Verified:
- Shutdown logs are printed as expected.
- nftables rules are properly removed.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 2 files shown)


aojea:
/ok-to-test
/lgtm
/approve

Thanks

PR #197, opened 2025-08-12
Introduce modular PolicyEvaluator architecture for better testability and extensibility

aojea:
This PR introduces a major refactor to compartmentalize and modularize the codebase by implementing a pipeline architecture and abstracting pod information into a single object. The purpose is to simplify integration of new AdminPolicy API changes and IP caches for large scale deployments.

Key changes include:

- Introduction of a new pipeline-based policy evaluation engine that supports multiple policy types (PolicyEvaluators) in a modular fashion
- Creation of a unified PodInfo abstraction that encapsulates pod, namespace, and node information, so PolicyEvaluators gets the necessary Pod Information from the IP address directly from the PolicyEngine
- Implementation of multiple IP resolution strategies through the IPResolver interface (informer-based and NRI-based), it is also possible to pass a custom resolver to get the Pod information from the IP address

Fixes: https://github.com/kubernetes-sigs/kube-network-policies/issues/121

The packet interception now can use its own controller with whatever technology, meanwhile it uses the PolicyEngine to evaluate packets


npinaeva (commented):
Very nice API, i think it makes sense

PR #199, opened 2025-08-15
Cluster Network Policy

aojea:
Implementation of the new v1alpha2 api for ClusterNetworkPolicy

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	k8s.io/component-helpers v0.33.4
 	k8s.io/klog/v2 v2.130.1
 	k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
-	sigs.k8s.io/network-policy-api v0.1.6-0.20250401132235-45061d10895e
+	sigs.k8s.io/network-policy-api v0.1.8-0.20250826165010-d5cee02b3c1c
 )
 
 require (
--- a/plugins/npa-v1alpha2/clusternetworkpolicy.go
+++ b/plugins/npa-v1alpha2/clusternetworkpolicy.go
@@ -1,0 +1,273 @@
+// SPDX-License-Identifier: APACHE-2.0
+
+package networkpolicy
+
+import (
+	"cmp"
+	"context"
+	"net"
+	"net/netip"
+	"slices"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/kube-network-policies/pkg/api"
+	"sigs.k8s.io/kube-network-policies/pkg/network"
+	"sigs.k8s.io/kube-network-policies/pkg/networkpolicy"
+	npav1alpha2 "sigs.k8s.io/network-policy-api/apis/v1alpha2"
+	cnpinformers "sigs.k8s.io/network-policy-api/pkg/client/informers/externalversions/apis/v1alpha2"
+	cnplisters "sigs.k8s.io/network-policy-api/pkg/client/listers/apis/v1alpha2"
+)
+
+// ClusterNetworkPolicy implements the PolicyEvaluator interface for the CNP API.
+type ClusterNetworkPolicy struct {
+	tier           npav1alpha2.Tier
+	cnpLister      cnplisters.ClusterNetworkPolicyLister
+	cnpSynced      cache.InformerSynced
+	domainResolver api.DomainResolver
+}
+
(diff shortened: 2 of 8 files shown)


pkg/networkpolicy/clusternetworkpolicy_test.go
```diff
@@ -0,0 +1,262 @@
+// SPDX-License-Identifier: APACHE-2.0
+
+package networkpolicy
```
aojea, line 3: generated with ai so this needs review

aojea:
/assign @danwinship @npinaeva @bowei 

I just vibe coded this in 30 mins to unblock the progress on the feature and conformance, so I sstill need to review the code too, but with the new architecture is pretty easy also to add unit tests

cmd/main.go
```diff
@@ -233,6 +258,14 @@ func run() int {
 		))
 	}
 
+	if clusterNetworkPolicy {
+		evaluators = append(evaluators, networkpolicy.NewClusterNetworkPolicy(
+			npav1alpha2.BaselineTier,
+			cnpInformer,
+			nil,
```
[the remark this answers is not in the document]
aojea: updated

cmd/main.go
```diff
@@ -131,7 +136,8 @@ func run() int {
 	var npaClient *npaclient.Clientset
 	var npaInformerFactory npainformers.SharedInformerFactory
 	var nodeInformer coreinformers.NodeInformer
-	if adminNetworkPolicy || baselineAdminNetworkPolicy {
+
+	if adminNetworkPolicy || baselineAdminNetworkPolicy || clusterNetworkPolicy {
```
npinaeva, line 140: do you want to check somewhere that ANP or BANP shouldn't be enabled together with CNP?

pkg/networkpolicy/clusternetworkpolicy.go
```diff
@@ -0,0 +1,253 @@
...
+type ClusterNetworkPolicy struct {
+	tier           npav1alpha2.Tier
+	cnpLister      cnplisters.ClusterNetworkPolicyLister
+	domainResolver DomainResolver
+}
+
+var _ PolicyEvaluator = &ClusterNetworkPolicy{}
+
+// NewClusterNetworkPolicy creates a new CNP implementation.
+func NewClusterNetworkPolicy(tier npav1alpha2.Tier, cnpInformer cnpinformers.ClusterNetworkPolicyInformer, domainResolver DomainResolver) *ClusterNetworkPolicy {
```
aojea, line 31: generated with AI so this also needs review

cmd/main.go
aojea: Should we? I think it makes sense to be One of as both sounds exclusive

aojea:
> @npinaeva this does not work as it is because the controller only matches on the pods that matches a network policy, you can force that by setting admin network policy the boolean to true on the controller config

fixed in last commit and image. uploaded to `aojea/kube-network-policies:cnp `

cmd/main.go
aojea: I will do it now

go.mod
```diff
@@ -114,3 +114,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.7.0 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace sigs.k8s.io/network-policy-api => github.com/npinaeva/network-policy-api v0.0.0-20250813121945-59370c2f182f
```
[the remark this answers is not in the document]
npinaeva: yes, it should work!
ping me when before merging it though, so that I can test the latest version locally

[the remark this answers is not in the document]
aojea: done, rebased

plugins/npa-v1alpha2/clusternetworkpolicy.go
```diff
@@ -0,0 +1,273 @@
...
+					_, cidr, err := net.ParseCIDR(string(network))
+					if err != nil {
+						continue
+					}
+					if cidr.Contains(dstIP) {
+						return rule.Action
+					}
+				}
+				for _, domain := range to.DomainNames {
+					if c.domainResolver != nil && c.domainResolver.ContainsIP(string(domain), dstIP) {
```
npinaeva, line 177: I think domainResolver is only nil for unit tests, but should we log something anyway in this case?

npinaeva (commented):
this code looks surprisingly good :D

npinaeva:
tested again, still works!

aojea:
/hold

wait for merging the conformance tests so we also add the ci jobs for the new apis

npinaeva:
let's get this in, then update the CI when the new conformance is merged
/lgtm

PR #217, opened 2025-08-19
[RFC][POC][WIP] Multi Cluster Network Policy Flat model

aojea:
Implement flat model of network policies across clusters

Ref: https://github.com/kubernetes-sigs/network-policy-api/issues/311

### DIFF
--- a/docs/multicluster.md
+++ b/docs/multicluster.md
@@ -1,0 +1,169 @@
+# Multi-Cluster Network Policies User Guide
+
+This guide provides instructions on how to install and use the multi-cluster
+feature of `kube-network-policies`. This feature allows you to enforce network
+policies across multiple Kubernetes clusters.
+
+### Overview
+
+The multi-cluster feature is composed of two main components:
+
+*   **`kube-ip-tracker`**: A central component that runs as a **Deployment** in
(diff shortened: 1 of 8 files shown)


docs/multicluster.md
```diff
@@ -0,0 +1,169 @@
...
+This guide provides instructions on how to install and use the multi-cluster
+feature of `kube-network-policies`. This feature allows you to enforce network
+policies across multiple Kubernetes clusters.
+
+### Overview
+
+The multi-cluster feature is composed of two main components:
+
+*   **`kube-ip-tracker`**: A central component that runs as a **Deployment** in
+    each cluster. It watches for `Pods`, `Namespaces`, and `Nodes` in all
```
skitt, line 12: Given how it works, it seems it could technically run in only one cluster (assuming a flat network), with the associated concerns regarding availability of course — but an unreachable IP tracker would imply an unreachable cluster so policy decidability at that point would be moot.

aojea: Yeah, but the recommended options is that you run a deployment with a `kube-ip-tracer` in EACH cluster.
This creates a full mesh and this guarantee that the local network policies depend of a component that runs in its own cluster.

PR #226, opened 2025-08-28
Fix push for npa-v1alpha2 as it was only building before

npinaeva:
You can see e.g. here https://storage.googleapis.com/kubernetes-ci-logs/logs/post-kube-network-policies-image/1960723853059559424/build-log.txt that there is logs for
`pushing manifest for gcr.io/k8s-staging-networking/kube-network-policies:v20250827-fa24992-npa-v1alpha1`
but no logs for `v1alpha2`


aojea:
@npinaeva we collided, please rebase to fix the conflict and we merge

Issue #246, opened 2025-09-20
re-evaluate active connections when policies change


aojea:
Full discussion in https://github.com/kubernetes-sigs/network-policy-api/issues/305

/cc @thockin @danwinship 

A simple thing will just to check the network policy diff on the create/update/delete and flush conntrack entries for the impacted connections/pods, but I'm worried about the side effect off flushing conntrack entries and wonder if there is something less disruptive in netfilter

aojea:
For UDP and SCTP we can just flush the conntrack entries, the new packet will hit the nfqueue and will be reevaluated.
For TCP conntrack when using `nf_conntrack_tcp_loose` not zero, that is the default,  `conntrack` is able to identify the `connection` and reestablish the conntrack entry. For this, we use SOCK_DESTROY https://lwn.net/Articles/666592/ , that shutsdown the socket associated to the connection,

aojea:
Patch to add a selftest to the kernel with this behavior on the tcp entry timeout https://patchwork.ozlabs.org/project/netfilter-devel/patch/20251020200805.298670-1-aojea@google.com/ , I do not know if is something the netfilter community will want to guarantee in the future so is better to discuss with them

PR #253, opened 2025-09-24
recalculate network policies for established connections

aojea:
Fixes: #246

### DIFF
--- a/pkg/dataplane/controller.go
+++ b/pkg/dataplane/controller.go
@@ -2,6 +2,7 @@ package dataplane
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -10,15 +11,21 @@ import (
 	"github.com/google/nftables/binaryutil"
 	"github.com/google/nftables/expr"
 	"github.com/mdlayher/netlink"
+	vishnetlink "github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
 	"golang.org/x/sys/unix"
 
+	v1 "k8s.io/api/core/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 
+	"sigs.k8s.io/kube-network-policies/pkg/dataplane/conntrack"
 	"sigs.k8s.io/kube-network-policies/pkg/network"
 	"sigs.k8s.io/kube-network-policies/pkg/networkpolicy"
 	"sigs.k8s.io/kube-network-policies/pkg/runner"
+	"sigs.k8s.io/kube-network-policies/pkg/socket"
 )
 
 // Network policies are hard to implement efficiently, and in large clusters this is
@@ -99,9 +106,22 @@ func newController(
 		1*time.Hour,   // maxInterval
 	)
 
+	// The conntrack cleaner will run periodically to remove entries
+	// that might have been left behind after network policies changed.
+	// The deletion of the entries will cause the connections to be dropped
+	// and re-evaluated by the new network policies.
+	c.conntrackReconciler = runner.NewBoundedFrequencyRunner(
+		controllerName+"-conntrack",
+		func() error { return c.firewallEnforcer(context.Background()) },
+		60*time.Second, // minInterval is conservative to avoid too much load on conntrack
+		10*time.Second, // retryInterval
+		1*time.Hour,    // maxInterval
+	)
+
 	// The sync callback now triggers the runner.
 	syncCallback := func() {
 		c.syncRunner.Run()
+		c.conntrackReconciler.Run()
 	}
 	c.policyEngine.SetDataplaneSyncCallbacks(syncCallback)
 
@@ -110,9 +130,10 @@ func newController(
 
 // Controller manages selector-based networkpolicy endpoints.
 type Controller struct {
-	config       Config
-	policyEngine *networkpolicy.PolicyEngine
-	syncRunner   *runner.BoundedFrequencyRunner
+	config              Config
+	policyEngine        *networkpolicy.PolicyEngine
+	syncRunner          *runner.BoundedFrequencyRunner
+	conntrackReconciler *runner.BoundedFrequencyRunner
 
 	nfq     *nfqueue.Nfqueue
 	flushed bool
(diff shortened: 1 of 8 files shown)


pkg/dataplane/controller.go
```diff
@@ -310,6 +329,98 @@ func (c *Controller) evaluatePacket(ctx context.Context, p *network.Packet) bool
...
+	}
+	var errorList []error
+	if len(filter.flowsV4) > 0 {
+		logger.V(2).Info("Cleaning IPv4 conntrack entries", "entries", len(filter.flowsV4))
+		n, err := vishnetlink.ConntrackDeleteFilters(vishnetlink.ConntrackTable, unix.AF_INET, filter)
+		if err != nil {
+			errorList = append(errorList, err)
+		}
+		logger.V(4).Info("Deleted conntrack entries", "entries", n)
+	}
```
danwinship, line 409: It's weird how `filter.addFlow()` hides IPv4-vs-IPv6 but then you have to think about it here... It seems like it would make more sense to move more of this into `conntrack.go`

pkg/dataplane/controller.go
```diff
@@ -310,6 +329,98 @@ func (c *Controller) evaluatePacket(ctx context.Context, p *network.Packet) bool
...
+		ipset.Insert(ip.String())
+	}
+
+	filter := newConntrackFilter()
+
+	for _, flow := range flows {
+		// TODO: only evaluate the flows that are established
+		// TCP for now, as they are the only ones supported by the library.
+		// UDP is not supported yet, but it can be done by parsing the raw attributes
+		// new flows will be evaluated by the nfqueue logic.
```
danwinship, line 365: I'm not sure exactly what this is saying. It needs better punctuation and capitalization and grammar...

"TODO: only evaluate the flows that are established" sounds like you're saying "in the future, we should only evaluate the flows that are established", except then it seems like you're saying "currently we only evaluate established TCP flows but in the future we should evaluate established UDP flows too".

Also not clear whether "not supported ... it can be done" means "the library could be extended to do this" or "we could do this ourselves without support from the library if we needed to".

Also I understand what the last sentence means, but it doesn't seem like it belongs _here_.

pkg/dataplane/controller.go
```diff
@@ -310,6 +329,98 @@ func (c *Controller) evaluatePacket(ctx context.Context, p *network.Packet) bool
...
+			continue
+		}
+		if state.State != nl.TCP_CONNTRACK_ESTABLISHED {
+			continue
+		}
+		// If divertAll is true, all pod IPs are managed by network policies.
+		if !divertAll {
+			// Only evaluate flows that are affected by network policies.
+			// Translated packets will be evaluated on the destination node.
+			if !ipset.Has(flow.Forward.SrcIP.String()) || !ipset.Has(flow.Forward.DstIP.String()) {
```
danwinship, line 380: (need to get vishvananda/netlink ported to `netip`...)

aojea: good luck :)

aojea: hmm, my bad, it seems the project is revamping and there are more maintainers today

https://github.com/vishvananda/netlink/pulls?q=is%3Apr+is%3Aclosed

aojea:
it does not work flushing the conntrack entry, the connection is not sent through the queue, subsequent packets are treated as established

pkg/dataplane/controller.go
aojea: conntrafk flush does not work, conntrack has some memory that recovers the state :/

pkg/dataplane/controller_test.go
```diff
@@ -274,6 +282,11 @@ table inet kube-network-policies {
 		ip6 saddr @podips-v6 queue flags bypass to 102
 		ip6 daddr @podips-v6 queue flags bypass to 102
 	}
+	chain forward-established {
+		type filter hook forward priority filter - 5; policy accept;
+		ct state established ip saddr . th sport . ip daddr . th dport . meta l4proto @block-tuple-v4 reject with icmp port-unreachable
+		ct state established ip6 saddr . th sport . ip6 daddr . th dport . meta l4proto @block-tuple-v6 reject with icmpv6 addr-unreachable
+	}
```
aojea, line 289: we need to do something like this , is the only way I find out we can "break" existing connections https://lore.kernel.org/netfilter-devel/20250313231341.3040002-1-aojea@google.com/T/#m44c410cd56bdc62cc717798a0e28ee4dcaaab9ad

aojea:
Ok, I have a better understanding now of the problem, assume TCP and UDP only by now

For UDP just clean up the conntrack entry, next one will be sent to userspace with the nfqueu
For TCP, things get complicated here as there area a lot of `ifs` and deleting the conntrack entry directly does not work
- if `nf_conntrack_tcp_loose` is enabled (default), conntrack picks  up already established connections, so current logic to only evaluate the first packet we'll not work if we just flush the conntrack entry because it will be considered ESTABLISHED. Conntrack stores the tcp sequence numbers and windows to understand if a packet belongs to the connection
- Good old TCP RST attack as in https://man.archlinux.org/man/extra/dsniff/tcpkill.8.en, it will require for us to have the sequence number and window to forge a packet, storing all the information is a no go, but we can try to get it from the conntrack , no idea if is already exposed or there is some ebpf magic that can return it, but injecting packets seems very intrusive.
- Add nftables reject rules per tuple, but this does not look like scales well and also does not handle connections that are idle, so we do have to monitor the conntrack entry to understand for how long we should leave the nftables rule to reject that connection
- sock_destroy , https://lwn.net/Articles/666220/ , this seems to be exactly what we are looking for,  `ss -K` already implements it., the challenge will be to find the socket, and it is also exposed via netlink,

https://lwn.net/Articles/1037388/
https://lpc.events/event/16/contributions/1358/attachments/1015/2101/socket-termination-lpc2022.pdf
https://docs.ebpf.io/linux/kfuncs/bpf_sock_destroy/



It seems cilium folks went through this already and have pretty good description of the problem and are adding the necessary tooling to the stack https://github.com/cilium/cilium/pull/25169

aojea:
> EDIT: oh, wait, that doesn't matter for NetworkPolicy purposes; you can postpone evaluation until the destination node

yeah, since we want to break existing connections, and it seems we need to be "invasive" we need to make local decisions only ... current code can optimize making decisions for the flow in both directions in the originating node ...

By the way, talked with one team mate working on the ebpf socket destroy problem internally, and he told me that there is no better solution than iterating through namespaces to find the sockets ... I need to check if we can get the information from the conntrack entry or we can store the relation Pod/IP to network namespace and use it, then the good news is that we can just use netlink, is what `ss -K` uses

pkg/dataplane/controller.go
aojea: autogenerated stuff, agents complete sentences and I give up fighting at the second line

PR #257, opened 2025-10-08
add Network Policy API v1alpha2

aojea:
add new v1alpha2 CI


danwinship:
/lgtm

I guess... There's an argument for "it would be good for k-n-p to still support v1alpha1", but there's not a very strong argument for "k-n-p should go to heroic efforts to support v1alpha1".

aojea:
let's try to keep both

danwinship:
/lgtm
if it all passes

PR #264, opened 2025-10-15
The Logging Network policy handlers is independent of the log level

aojea:
The intention is that the consumer uses the logging handler at the log level they please, by inserting it in the pipeline. Hence, the logging handler should not make any assumption about verbosity levels 

Spotted by @danwinship  in https://github.com/kubernetes-sigs/kindnet/pull/11#discussion_r2432484298


aojea:
It is better/easier to drop support https://github.com/kubernetes-sigs/kube-network-policies/pull/264 , it also requires to pin the versions of teh conformance tests, in addition to the CRD manifests ... this will unblock the CI failures

aojea:
this is no controversial , just remove the log level that was clear an oversight, merging

PR #266, opened 2025-10-15
pin anp and bnp crds manifests

aojea:
It got dropped upstream so we need to use a fixed commit that still has support.

This should fix CI


aojea:
/assign @danwinship 

let's get this passing, then I will try to get the alpha2 in parallel

PR #303, opened 2026-01-30
Update to the API changes for ClusterNetworkPolicy

bowei:


plugins/npa-v1alpha2/clusternetworkpolicy.go
```diff
@@ -218,46 +218,81 @@ func (c *ClusterNetworkPolicy) evaluateClusterIngress(
...
+					return true
+				}
+				if policyProtocol.TCP.DestinationPort.Range != nil &&
+					policyProtocol.TCP.DestinationPort.Range.Start <= int32(port) &&
+					policyProtocol.TCP.DestinationPort.Range.End >= int32(port) {
+					return true
+				}
+			} else {
+				// If DestinationPort is nil, it means all TCP ports are allowed.
+				return true
```
danwinship: If DestinationPort is nil, then the policy is invalid. In the future, if/when we add something like `flags`, this clause will still not be correct. So just remove this.

plugins/npa-v1alpha2/clusternetworkpolicy_test.go
```diff
@@ -146,36 +145,76 @@ func Test_ClusterNetworkPolicy_Evaluation(t *testing.T) {
...
+	adminDenyAllEgress := makeClusterNetworkPolicy("admin-deny-all-egress", 100, npav1alpha2.AdminTier, func(p *npav1alpha2.ClusterNetworkPolicy) {
+		p.Spec.Subject = npav1alpha2.ClusterNetworkPolicySubject{
+			Namespaces: &metav1.LabelSelector{},
+		}
+		p.Spec.Egress = []npav1alpha2.ClusterNetworkPolicyEgressRule{
+			{
+				Action: npav1alpha2.ClusterNetworkPolicyRuleActionDeny,
+				To: []npav1alpha2.ClusterNetworkPolicyEgressPeer{
+					{
+						Pods: &npav1alpha2.NamespacedPod{
```
danwinship: Any reason you changed this to a more verbose way of saying the same thing? (`Namespaces` vs `Pods`)
(Likewise several times below.)

(And unrelated cleanups should be a separate commit from porting Ports to Protocols.)

danwinship:
@bowei you need to update the policies in `tests/*.bats` and `./install-iptracker.yaml`

bowei:
We were doing some experiments and it seems like the SCTP tests pass when run in isolation. Trying to pin down why it is failing.

bowei:
It seems like increasing the timeout fixed the tests.

aojea:
great

PR #306, opened 2026-02-07
Set up .go-version file and bump to Go 1.25

stmcginnis:
This follows the existing pattern of using a .go-version file to denote the version of Go to be used. Workflows can then extract the version from that file and updates don't need to remember to modify every place where it is used.

This also raises the Go version to 1.25 to align with k/k and make it possible to update k8s.io dependencies for the versions that now require 1.25 or later.


stmcginnis:
Probably helps to check out the code before trying to read it...

PR #338, opened 2026-04-12
Add controller nfqueue integration test

nickgarlis:
Add TestController_Run to exercise the full nfqueue packet path: nftables rule sync, packet interception, verdict, and delivery.

Add skipSkuidBypass config option to disable the "meta skuid 0 accept" rule, allowing the test to force traffic through nfqueue from a user namespace where only UID 0 is mapped.

The test currently fails with mdlayher/netlink v1.11.0 due to a bug in netlink.Message.UnmarshalBinary rejecting valid unaligned netlink messages from nfqueue.

### DIFF
--- a/pkg/dataplane/controller.go
+++ b/pkg/dataplane/controller.go
@@ -55,6 +55,7 @@ type Config struct {
 	NFTableName         string // if other projects use this controllers they need to be able to use their own table name
 	StrictMode          bool   // enforce network policies also on established connections
 	CTLabelAccept       int    // conntrack label to set on accepted connections (between 1-127)
+	skipSkuidBypass     bool   // skip the "meta skuid 0 accept" rule, used for testing to force all traffic through nfqueue
 }
 
 func (c *Config) Defaults() error {
@@ -630,16 +631,18 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 	// or system daemons that depend on the internal node traffic to not be blocked.
 	// Ref: https://github.com/kubernetes-sigs/kube-network-policies/issues/65
 	// meta skuid 0 accept
-	nft.AddRule(&nftables.Rule{
-		Table: table,
-		Chain: chain,
-		Exprs: []expr.Any{
-			&expr.Meta{Key: expr.MetaKeySKUID, SourceRegister: false, Register: 0x1},
-			&expr.Cmp{Op: expr.CmpOpEq, Register: 0x1, Data: []byte{0x0, 0x0, 0x0, 0x0}},
-			&expr.Counter{},
-			&expr.Verdict{Kind: expr.VerdictAccept},
-		},
-	})
+	if !c.config.skipSkuidBypass {
+		nft.AddRule(&nftables.Rule{
+			Table: table,
+			Chain: chain,
+			Exprs: []expr.Any{
+				&expr.Meta{Key: expr.MetaKeySKUID, SourceRegister: false, Register: 0x1},
+				&expr.Cmp{Op: expr.CmpOpEq, Register: 0x1, Data: []byte{0x0, 0x0, 0x0, 0x0}},
+				&expr.Counter{},
+				&expr.Verdict{Kind: expr.VerdictAccept},
+			},
+		})
+	}
 
 	// The queue sets the conntrack mark for the packets it processes,
 	// so we can clear the mark here later to re-process connections if needed.
--- a/pkg/dataplane/controller_test.go
+++ b/pkg/dataplane/controller_test.go
@@ -2,6 +2,7 @@ package dataplane
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"net/netip"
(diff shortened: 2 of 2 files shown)


pkg/dataplane/controller_test.go
```diff
@@ -426,3 +435,87 @@ func compareMultilineStringsIgnoreIndentation(str1, str2 string) bool {
...
 	return str1 == str2
 }
+
+// TestController_Run exercises the full dataplane path: nftables rule sync,
+// nfqueue packet interception, verdict processing, and packet delivery.
+// It verifies that Controller.Run correctly receives and processes packets
+// through the nfqueue netlink socket.
+func TestController_Run(t *testing.T) {
+	if !unpriviledUserns() {
+		t.Skip("Test requires unprivileged user namespaces")
```
nickgarlis, line 515: I don't think it's easy to test this on root since a netns would only be used by one thread

pkg/dataplane/controller_test.go
```diff
@@ -426,3 +435,87 @@ func compareMultilineStringsIgnoreIndentation(str1, str2 string) bool {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- controller.Run(ctx)
+	}()
+
+	// Give the controller time to set up nftables rules and open nfqueue.
+	time.Sleep(500 * time.Millisecond)
```
nickgarlis: Not too elegant but polling may have been too much

pkg/dataplane/controller.go
```diff
@@ -55,6 +55,7 @@ type Config struct {
 	NFTableName         string // if other projects use this controllers they need to be able to use their own table name
 	StrictMode          bool   // enforce network policies also on established connections
 	CTLabelAccept       int    // conntrack label to set on accepted connections (between 1-127)
+	skipSkuidBypass     bool   // skip the "meta skuid 0 accept" rule, used for testing to force all traffic through nfqueue
```
nickgarlis, line 58: The problem is that the rule is matched if you're the owner of the user namespace. The alternative would have been to remove the rule at test-time to force the packet to nfqueue.

aojea: If the test target nfqueue specifically... Should not be better to add the test also on the go-nfqueue project, the maintainer is very responsive? I like this test and there is also a feature request to handle this specific rule better, so please bear with me a bit more ...

nickgarlis: The reason I wrote this test was for me to be able to figure out what the issue was without running the full e2e tests.

I’m not familiar with this repo, so I’m not sure this is the right approach. My goal was to keep the test generic rather than nfqueue-specific, but the bypass rule ended up exercising a path that isn’t typical since tests usually run as root or in a user namespace.

I’ve opened this PR in case the test is helpful to you too. Happy to fix the failing tests or revise the approach if you want to keep it.

aojea: it is helpful , really helpful and I want it, is just I want to see how we can structure it for adding more integration tests and not have this strong dependency on e2e test that are really hard to debug

nickgarlis:
On the surface, the failing test does not look related but I am not sure. Perhaps it's a flaky one ?

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xd8 pc=0x655cd99]
goroutine 430 [running]:
k8s.io/kops/pkg/resources/gce.(*clusterDiscoveryGCE).listHealthchecks.func1({0x7525d40?, 0x81a1a21?}, 0xc0008f82d0?)
	k8s.io/kops/pkg/resources/gce/gce.go:1144 +0x199
k8s.io/kops/pkg/resources/ops.DeleteResources.func1({0xc0003c2088, 0x1, 0x1})
	k8s.io/kops/pkg/resources/ops/delete.go:129 +0x37f
created by k8s.io/kops/pkg/resources/ops.DeleteResources in goroutine 1
	k8s.io/kops/pkg/resources/ops/delete.go:110 +0xc4f
Error: exit status 2
```

aojea:
> On the surface, the failing test does not look related but I am not sure. Perhaps it's a flaky one ?
> 
> ```
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0xd8 pc=0x655cd99]
> goroutine 430 [running]:
> k8s.io/kops/pkg/resources/gce.(*clusterDiscoveryGCE).listHealthchecks.func1({0x7525d40?, 0x81a1a21?}, 0xc0008f82d0?)
> 	k8s.io/kops/pkg/resources/gce/gce.go:1144 +0x199
> k8s.io/kops/pkg/resources/ops.DeleteResources.func1({0xc0003c2088, 0x1, 0x1})
> 	k8s.io/kops/pkg/resources/ops/delete.go:129 +0x37f
> created by k8s.io/kops/pkg/resources/ops.DeleteResources in goroutine 1
> 	k8s.io/kops/pkg/resources/ops/delete.go:110 +0xc4f
> Error: exit status 2
> ```

that is a kops stack trace, tagging kops maintainers for visibility @hakman @justinsb @rifelpet

pkg/dataplane/controller_test.go
aojea: we can take a different approach, you are using allow only rules, if you define allow and deny rules, that can give you the right signal to know that packets are being processed.

If you set allow only to port 12345 and deny to port 54321, you just send X packets to port 54321 until it is denied or dropped. Then you send traffic to port 12345 that has to go through

pkg/dataplane/controller_test.go
```diff
@@ -426,3 +436,87 @@ func compareMultilineStringsIgnoreIndentation(str1, str2 string) bool {
...
+
+	// Verify the controller is still running
+	select {
+	case err := <-errCh:
+		t.Fatalf("controller.Run exited prematurely: %v", err)
+	default:
+	}
+
+	listener, err := net.ListenUDP("udp", &net.UDPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 0})
```
aojea: just curious, why UDP?

nickgarlis: It seemed like the simplest approach since we are only observing the packet flow and are sending one packet. If we wanted to test stateful connections where conntrack rules are used then I guess TCP would also make sense ?

rifelpet:
> that is a kops stack trace, tagging kops maintainers for visibility

This should be fixed now

pkg/dataplane/controller_test.go
aojea: drive by comment, udp use to be more flaky because there is no guarantee of delivery, TCP guarantees that make test reliable 100%

nickgarlis: right, I was not appreciating the fact that delivery is not guaranteed on loopback. I'll switch it to TCP :+1:

nickgarlis:
@aojea I've pushed a new commit with the TCP approach but while it does seem to work well, it does not fail for that particular bug. The reason is that during the handshake phase the first SYN packet is 4 byte aligned and we then get into the `ct established,related` path for any subsequent packets of the same connection.

However, this is arguably a better test for your repository since it tests all paths.

Another approach could be to go back to UDP and simply send messages in a loop until one of them is received or the test times out. Or perhaps both tests (TCP & UDP) would be useful.

aojea:
> Or perhaps both tests (TCP & UDP) would be useful.

I like if you can add both, so one servers as correlation of the other

pkg/dataplane/controller.go
aojea: resolving, this looks fine and is an internal detail

aojea:
/lgtm
/approve

I left the UDP test as a future improvement,

Thanks

PR #339, opened 2026-04-14
implement egress policies from Pod to Node

### LINKED ISSUE #283: Same node traffic is always allowed not only in ingress, but also in egress direction
With the default-deny behaviour of network policy we had to make sure kubelet probes continue to work, hence we made an exception for the same-node-to-pod ingress connection to be always allowed. The docs say https://kubernetes.io/docs/concepts/services-networking/network-policies/#the-two-sorts-of-pod-isolation
```
When a pod is isolated for egress, the only allowed connections from the pod are those allowed by the egress list of some NetworkPolicy that applies to the pod for egress.
...
When a pod is isolated for ingress, the only allowed connections into the pod are those from the pod's node and those allowed by the ingress list of some NetworkPolicy that applies to the pod for ingress.
```
Note how only ingress direction mentions `the pod's node` to be allowed.
I stumbled on this when running pod-to-host tests in CNP and noticed that pod-to-same-node is always allowed.

aojea:
We were evaluating traffic only on the POSTROUTING hook. However, we also need to evaluate packets going from Pods to Nodes.

Instead of moving policy evaluation to PREROUTING, this change extends the current logic by adding INPUT hook processing to inspect traffic that:

is not host-originated,
is in ct state NEW, and
is directed to the host itself

Fixes: #283

### DIFF
--- a/pkg/dataplane/controller.go
+++ b/pkg/dataplane/controller.go
@@ -737,6 +737,75 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 		},
 	})
 
+	// INPUT chain: evaluate pod-to-host (egress toward the node itself) network policies.
+	// Only the first packet of new connections (ct state new) is queued.
+	// TODO: investigate why if we process not only new, e2e tests start failing.
+	inputChain := nft.AddChain(&nftables.Chain{
+		Name:    "input",
+		Table:   table,
+		Type:    nftables.ChainTypeFilter,
+		Hooknum: nftables.ChainHookInput,
+		// Run after IPVS LOCAL_IN hooks (which use srcnat-2 and srcnat-1 priorities).
+		// https://elixir.bootlin.com/linux/v5.10/source/net/netfilter/ipvs/ip_vs_core.c#L2249
+		Priority: nftables.ChainPriorityRef(*nftables.ChainPriorityNATSource + 1),
+	})
+
+	// iifname "lo" accept - bypass all loopback traffic to avoid blocking node-to-itself traffic.
+	nft.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: inputChain,
+		Exprs: []expr.Any{
+			&expr.Meta{Key: expr.MetaKeyIIFNAME, SourceRegister: false, Register: 0x1},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 0x1, Data: []byte("lo\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")},
+			&expr.Verdict{Kind: expr.VerdictAccept},
+		},
+	})
+
+	if !divertAll {
+		// ip saddr @podips-v4 ct state new queue
+		nft.AddRule(&nftables.Rule{
+			Table: table,
+			Chain: inputChain,
+			Exprs: []expr.Any{
+				&expr.Meta{Key: expr.MetaKeyNFPROTO, SourceRegister: false, Register: 0x1},
+				&expr.Cmp{Op: expr.CmpOpEq, Register: 0x1, Data: []uint8{unix.NFPROTO_IPV4}},
+				&expr.Payload{DestRegister: 0x1, Base: expr.PayloadBaseNetworkHeader, Offset: 12, Len: 4},
+				&expr.Lookup{SourceRegister: 0x1, DestRegister: 0x0, SetName: "podips-v4"},
+				&expr.Ct{Register: 0x1, SourceRegister: false, Key: expr.CtKeySTATE},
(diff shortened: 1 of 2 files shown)


aojea:
Ok, ipvs fails, that is consistent with the history on the repo and reinforce the idea yo avoid pre routing

danwinship:
e2e still failing...

npinaeva:
ip_tracker though?
I will test this tomorrow, hoping e2e just got unlucky

aojea:
it is passing , the failure is an unrelated flake

```
2026-04-14T15:11:45.0493940Z             - lastProbeTime: null
2026-04-14T15:11:45.0494392Z               lastTransitionTime: "2026-04-14T15:06:04Z"
2026-04-14T15:11:45.0495541Z               message: '0/3 nodes are available: 1 node(s) had untolerated taint {node-role.kubernetes.io/control-plane:
2026-04-14T15:11:45.0496659Z                 }, 2 Too many pods. preemption: 0/3 nodes are available: 1 Preemption is not
2026-04-14T15:11:45.0497580Z                 helpful for scheduling, 2 No preemption victims found for incoming pod.'
2026-04-14T15:11:45.0498253Z               reason: Unschedulable
```

aojea:
> e2e still failing...

All green now 😄

pkg/dataplane/controller.go
```diff
@@ -737,6 +737,74 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
+				&expr.Payload{DestRegister: 0x1, Base: expr.PayloadBaseNetworkHeader, Offset: 8, Len: 16},
+				&expr.Lookup{SourceRegister: 0x1, DestRegister: 0x0, SetName: "podips-v6"},
+				&expr.Ct{Register: 0x1, SourceRegister: false, Key: expr.CtKeySTATE},
+				&expr.Bitwise{SourceRegister: 0x1, DestRegister: 0x1, Len: 0x4, Mask: binaryutil.NativeEndian.PutUint32(expr.CtStateBitNEW), Xor: []byte{0x0, 0x0, 0x0, 0x0}},
+				&expr.Cmp{Op: expr.CmpOpNeq, Register: 0x1, Data: []byte{0x0, 0x0, 0x0, 0x0}},
+				queue,
+			},
+		})
+	} else {
+		// ct state new queue
```
npinaeva: one thing about ct state handling: the postrouting rules handle new vs related connections a bit differently:
```
ct label 28 ct state established,related counter packets 250 bytes 21000 accept
queue flags bypass to 89
ct label set 28
```
is there a reason to express it differently for the input hook?

npinaeva:
Experimental CNP conformance passes when used together with https://github.com/kubernetes-sigs/kube-network-policies/pull/340! so it works :P

aojea:
well is on you on approving the PR

pkg/dataplane/controller.go
aojea: oh, right, we implemented the logic to break the existing connections if the network policies change

npinaeva: so do we want to have the same for input?

npinaeva:
> well is on you on approving the PR

hehe sure, I am waiting to resolve the only comment I had :)

pkg/dataplane/controller.go
aojea: yeah, we should

aojea: done

aojea:
> > well is on you on approving the PR
> 
> hehe sure, I am waiting to resolve the only comment I had :)

updated

pkg/dataplane/controller.go
```diff
@@ -740,6 +737,100 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
+		nft.AddRule(&nftables.Rule{
+			Table: table,
+			Chain: inputChain,
+			Exprs: []expr.Any{
+				&expr.Ct{Register: 0x1, SourceRegister: false, Key: expr.CtKeySTATE},
+				&expr.Bitwise{SourceRegister: 0x1, DestRegister: 0x1, Len: 0x4, Mask: binaryutil.NativeEndian.PutUint32(expr.CtStateBitNEW), Xor: []byte{0x0, 0x0, 0x0, 0x0}},
```
npinaeva: so do you want to keep a match on new here considering there is now a related match on top of it? Asking because it it still a bit different from the `postrouting` and may be confusing in the future

aojea: me being lazy, updated to be consisten

aojea: is this comment related to this chain? I think was realted to the chain and got changed with the latest push

aojea:
hmm, it seems removing the `new` statement is causing some undesired problem with ipv6

aojea:
@npinaeva I could not investigate more but removing the `ct state new` breaks ipv6 jobs, I think current state is good to go, despite different is safer, since is traffic directed to the host that is specially sensible, and conntrack state table is super complex, so being explicit here seems the right call to me

aojea:
> /approve

@npinaeva you need to provide LGTM :)

PR #340, opened 2026-04-15
Fix node selectors for CNP

npinaeva:
Field doc is here https://github.com/kubernetes-sigs/network-policy-api/blob/72895f2a4a97653784126b29a6f32fbb631b708f/apis/v1alpha2/clusternetworkpolicy_types.go#L363-L368 explaining how it should work
Also enable experimental conformance test.
Depends on https://github.com/kubernetes-sigs/kube-network-policies/pull/339

### DIFF
--- a/plugins/npa-v1alpha2/clusternetworkpolicy.go
+++ b/plugins/npa-v1alpha2/clusternetworkpolicy.go
@@ -10,7 +10,9 @@ import (
 	"slices"
 
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/kube-network-policies/pkg/api"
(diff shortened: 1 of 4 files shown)


aojea:
cool, let's get https://github.com/kubernetes-sigs/kube-network-policies/pull/339 in and then rebase

aojea:
I'm not familiar with the expected semantics, @danwinship and/or @tssurya can you please review the PR?

plugins/npa-v1alpha2/clusternetworkpolicy.go
```diff
@@ -160,8 +165,26 @@ func (c *ClusterNetworkPolicy) evaluateClusterEgress(
...
+						continue
+					}
+					nodes, err := c.nodeLister.List(s)
+					if err != nil {
+						continue
+					}
+					for _, node := range nodes {
+						for _, addr := range node.Status.Addresses {
+							if addr.Type == v1.NodeInternalIP || addr.Type == v1.NodeExternalIP {
+								ip := net.ParseIP(addr.Address)
```
danwinship, line 180: We should cache this info rather than re-parsing it on every potentially-matching packet.
Given that @aojea already approved this, we don't need to do this _now_, but can you file an issue suggesting someone do this?

(Note that if you use `netip.Addr` instead of `net.IP` you can create a `sets.Set[netip.Addr]` for each node and then just check if `nodeAddrs.Has(dstAddr)`)

PR #369, opened 2026-07-04
Add portable image promotion tool and make target

aojea:


aojea:
no code impacted ,just tooling for contributors able to help with image promotion
/cc @stmcginnis @michaelasp

aojea:
still missing the update of the current repo manifests

aojea:
ok, this is a nobrainer, so easy now, I can do the release in 1 min, before it was 3 different steps

PR #381, opened 2026-07-21
networkpolicy: exclude hostNetwork pods from managed IPs

### LINKED ISSUE #380: ManagedIPs() enrolls hostNetwork pods' node IPs into the podips-v4/v6 enforcement set (no HostNetwork skip, unlike podinfo.ipresolver)
## Environment

- kube-network-policies **v1.0.0** (`registry.k8s.io/networking/kube-network-policies:v1.0.0`), deployed as a sidecar in the flannel DaemonSet (Talos Linux, `kubeNetworkPoliciesEnabled: true`)
- Kubernetes v1.36.x, kube-proxy `--proxy-mode=iptables`
- Flannel (vxlan), 12+ node bare-metal cluster (mixed control-plane + worker roles on the same physical nodes in some cases)

## What happened

`StandardNetworkPolicy.ManagedIPs()` builds the nftables `podips-v4`/`podips-v6` enforcement address set from every pod selected (locally, per-node) by any `NetworkPolicy`, with **no exclusion for `hostNetwork: true` pods**:

https://github.com/kubernetes-sigs/kube-network-policies/blob/v1.0.0/pkg/networkpolicy/networkpolicy.go#L140-L160

```go
func (s *StandardNetworkPolicy) ManagedIPs(ctx context.Context) ([]netip.Addr, bool, error) {
	allPolicies, err := s.networkpolicyLister.List(labels.Everything())
	...
	for _, policy := range allPolicies {
		pods := s.getLocalPodsForNetworkPolicy(policy)
		for _, pod := range pods {
			for _, podIP := range pod.Status.PodIPs {
				addr, err := netip.ParseAddr(podIP.IP)
				if err == nil {
					managedIPs.Insert(addr)
				}
			}
		}
	}
	...
}
```

`getLocalPodsForNetworkPolicy` (same file, `L163-183`) likewise applies no `HostNetwork` filter — it only filters by selector match + local node.

For a `hostNetwork: true` pod (e.g. a static `kube-apiserver` pod, or any DaemonSet pod needing host networking such as a media-plane VoIP workload, a MetalLB speaker, a Prometheus node-exporter, or a custom hostNetwork workload), `pod.Status.PodIPs` **is the node's own IP**. Any `NetworkPolicy` whose selector matches that pod — including a namespace's common empty-selector (`podSelector: {}`) "default-deny"/"default-allow" baseline policies, which is an extremely common pattern — inserts the **node's real IP** into `podips-v4`.

Once a node's own IP is in that set, the nfqueue rule (`ip daddr @podips-v4 ... queue ... to 100`, `flags bypass`) matches **every packet whose source or destination is that node IP**, not just packets belonging to the offending pod. Observed consequences on our cluster:

1. **Same-node pod → own-node apiserver drops.** A pod on a node that also runs a control-plane `kube-apiserver` static pod (`hostNetwork: true`, matched by an empty-selector `NetworkPolicy` in `kube-system`) fails to reach `https://<own-node-IP>:6443` — because that node IP is now in `podips-v4` and gets queued/evaluated as if it were pod traffic, then denied for lack of a matching rule (this failure mode is a *consequence* of the address-set pollution described here, though the actual drop mechanics for that specific hairpin/reply case overlap with #379 — see below).
2. **All VXLAN encapsulation traffic gets nfqueued.** Once a node's IP is a "managed" address, *all* UDP/4789 (VXLAN) traffic to/from that node — i.e. essentially all inter-node pod traffic transiting through it — gets pulled through the userspace evaluation path, producing tens of thousands of nfqueue evaluations per 15s window on a moderately busy node and materially increasing CPU overhead for the sidecar.
3. **Misattributed drops.** Because the nfqueue verdict logs key off whichever local pod happens to share the flagged node IP's "managed" status contextually, drops were initially misattributed to an unrelated pod (`logging/loki-chunks-cache-0` in our case) rather than to the real cause (a hostNetwork pod's IP leaking into the set via an unrelated namespace's default policy).

## Why this is a bug, not intended behavour

- The Kubernetes NetworkPolicy spec documents that hostNetwork pods are not subject to NetworkPolicy enforcement (their network identity is the node's), regardless of any selector match. `ManagedIPs()` should therefore never add a hostNetwork pod's IP to the enforcement set — matching it for evaluation purposes when building the *dataplane's own dataplane-sync state* is meaningless (there is no pod-network identity to enforce), but the side effect of doing so is to enroll the **node's own IP** into a set meant to scope pod IPs, with the consequences above.
- The codebase already has the correct exclusion implemented elsewhere and simply isn't calling it from this path: `pkg/podinfo/ipresolver.go` L46 explicitly skips hostNetwork pods when indexing pod IPs for its (unrelated) IP→pod lookup resolver:

  https://github.com/kubernetes-sigs/kube-network-policies/blob/v1.0.0/pkg/podinfo/ipresolver.go#L46

  ```go
  PodIPIndex: func(obj interface{}) ([]string, error) {
      pod, ok := obj.(*v1.Pod)
      if !ok || pod.Spec.HostNetwork {
          return []string{}, nil
      }
      ...
  },
  ```

  `ManagedIPs()` never reuses this check (it reads `pod.Status.PodIPs` directly off the lister result of `getLocalPodsForNetworkPolicy`, which has no HostNetwork filter at all).

## Impact / who hits this

Any cluster with a common, otherwise-correct pattern of "baseline" `NetworkPolicy` objects with an empty or broad `podSelector` (e.g. `default-deny-ingress`, `default-egress`) co-resident in a namespace with **any** hostNetwork pod — static control-plane pods in `kube-system`, MetalLB speakers, Prometheus node-exporter, or any hostNetwork DaemonSet workload — will silently enroll that node's IP into the enforcement address set, with the traffic-hairpinning/overhead/misattribution symptoms above. We found and had to work around this in 5 namespaces on one cluster (`kube-system`, `metallb-system`, `monitoring`, plus two application namespaces running hostNetwork workloads).

## Suggested fix

In `ManagedIPs()` (or in `getLocalPodsForNetworkPolicy`), skip pods with `pod.Spec.HostNetwork == true`, consistent with the exclusion already present in `pkg/podinfo/ipresolver.go`.

## Workaround in use

Until this is fixed upstream, we've narrowed every empty/broad-selector `NetworkPolicy` co-resident with hostNetwork pods to explicitly exclude them via `matchExpressions` (`NotIn` on a label the hostNetwork pods carry — falling back to a self-contradictory `In`+`NotIn` pair on policies whose entire original match population was hostNetwork-only, to avoid widening the selector to unrelated pods). This is namespace/deployment-specific manual work that shouldn't be necessary given the spec's own documented hostNetwork exemption.

## Related but distinct

#379 documents a related-looking but **separate** bug in the same package/version: same-node hairpin **reply** packets (e.g. an apiserver SYN-ACK back to a pod on the same node) being evaluated as new external Ingress against the un-reverse-DNAT'd tuple and dropped for lack of a matching Ingress rule. We can independently confirm #379's exact symptom and log signature on our cluster (`"Evaluating packet" direction="Ingress" srcPod="external" dstPod=...` against the raw `nodeIP:6443 → podIP:ephemeral` tuple) as a **residual** issue that persists even after fully fixing the `ManagedIPs()` leak described here — i.e. these are two independent bugs that happen to both manifest around same-node/hairpin apiserver traffic on nodes that are both a workload node and a control-plane node, which is why we initially conflated them during triage. Filing this separately since the root cause and fix are unrelated to #379's.

ColumbusLabs:
/kind bug

## What

Exclude `hostNetwork` pods when selecting local pods managed by standard NetworkPolicies, and add a dual-stack regression test for `ManagedIPs`.

## Why

Host-networked pods use their node's IP address. Adding those addresses to the managed pod IP sets can divert and evaluate node traffic as pod traffic.

## Testing

- `CGO_ENABLED=1 go test -race -count=1 ./pkg/networkpolicy`
- `golangci-lint run ./pkg/networkpolicy`
- `gofmt -d pkg/networkpolicy/networkpolicy.go pkg/networkpolicy/networkpolicy_test.go`
- `git diff --check`

The full repository test target is Linux-specific and does not compile on macOS because `github.com/google/nftables/xt` references Linux-only `unix.NFPROTO_*` constants; the complete changed package race suite passes.

Fixes #380

### DIFF
--- a/pkg/networkpolicy/networkpolicy.go
+++ b/pkg/networkpolicy/networkpolicy.go
@@ -175,7 +175,7 @@ func (s *StandardNetworkPolicy) getLocalPodsForNetworkPolicy(networkPolicy *netw
 	}
 	result := []*v1.Pod{}
 	for _, pod := range pods {
-		if pod.Spec.NodeName == s.nodeName {
+		if pod.Spec.NodeName == s.nodeName && !pod.Spec.HostNetwork {
 			result = append(result, pod)
 		}
 	}
--- a/pkg/networkpolicy/networkpolicy_test.go
+++ b/pkg/networkpolicy/networkpolicy_test.go
@@ -9,6 +9,7 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/component-base/logs"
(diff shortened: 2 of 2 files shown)


aojea:
/ok-to-test
@ColumbusLabs this looks ok but you need to sign the CLA

aojea:
@ColumbusLabs let me know if you want to proceed or I will take over, we want the patch to fix the bug but policies does not allow contributing without signing the CLA

PR #403, opened 2026-09-19
Do not recreate the nftables base chains on every sync

### LINKED ISSUE #402: Packets waiting in an nfqueue are dropped every time the nftables table is rebuilt
**What happened**

The agent logs this, and packets that were waiting for a verdict never arrive:

```
"Could not receive message" error="netlink receive: no such file or directory"
```

It happens on every rules sync. Both `syncNFTablesRules` (`pkg/dataplane/controller.go`) and the dnscache's `syncRules` (`pkg/dns/dnsagent.go`) rebuild their table with `AddTable`, `DelTable`, `AddTable`. Deleting a table deletes its base chains, and deleting a base chain unregisters its netfilter hook. The kernel reacts by dropping every packet waiting in **any** nfqueue of that network namespace, so a dnscache sync drops packets waiting in the policy queue too.

In v6.18 the path is: `nft_flush_table` calls `nft_delchain` for each chain (`nf_tables_api.c:1653`), committing that calls `nf_tables_unregister_hook` (`:11032`), which reaches `nf_queue_nf_hook_drop(net)` (`core.c:522`), which calls `nfqnl_flush(inst, NULL, 0)` on every queue instance (`nfnetlink_queue.c:1184`), reinjecting each entry with `NF_DROP` (`:403`).

`fail-open` does not help here, because this is an explicit drop rather than a failure to enqueue.

The verdicts that arrive afterwards then match no entry, so the kernel returns `-ENOENT` (`nfnetlink_queue.c:1435`). It replies with that error even though `setVerdict` does not set `NLM_F_ACK`, because `nfnetlink.c:538` answers when the flag is set **or** an error occurred. Since the write path never reads a reply, the error surfaces asynchronously in the receive loop, which is why `SetVerdict` itself never returns an error.

The npa builds make this routine: the dnscache syncs every 30 seconds whether or not anything changed. The dataplane sync is event driven, so it lands during pod and policy churn.

**What you expected to happen**

A rules sync should not drop packets that are waiting for a verdict.

**How to reproduce it (as minimally and precisely as possible)**

No cluster needed. This uses the same nftables and nfqueue libraries as this repo. It queues UDP packets, delays each verdict by 50 ms so packets are waiting, and rebuilds the table every 500 ms for 10 seconds.

```go
// Shows that rebuilding an nftables table that has a hooked chain drops every
// packet waiting in an nfqueue.
//
//	-mode none        never rebuild, as a baseline
//	-mode recreate    rebuild with AddTable + DelTable + AddTable
//	-mode flushchain  keep the table, flush the chain and re-add the rule
//	-mode othertable  never touch the queue's table, rebuild an unrelated one
package main

import (
	"context"
	"errors"
	"flag"
	"fmt"
	"net"
	"os"
	"sync/atomic"
	"syscall"
	"time"

	"github.com/florianl/go-nfqueue/v2"
	"github.com/google/nftables"
	"github.com/google/nftables/expr"
)

const (
	queueNum = 100
	port     = 9999
	duration = 10 * time.Second
	hold     = 50 * time.Millisecond // verdict delay, so packets wait in the queue
	sendGap  = 2 * time.Millisecond
	syncGap  = 500 * time.Millisecond
)

var (
	table = &nftables.Table{Name: "flushtest", Family: nftables.TableFamilyINet}
	other = &nftables.Table{Name: "flushtest-other", Family: nftables.TableFamilyINet}
)

// rebuildOther recreates a second, unrelated table that has its own hooked chain
// and no queue rule at all.
func rebuildOther() error {
	c, err := nftables.New()
	if err != nil {
		return err
	}
	c.AddTable(other)
	c.DelTable(other)
	c.AddTable(other)
	c.AddChain(&nftables.Chain{
		Name: "output", Table: other, Type: nftables.ChainTypeFilter,
		Hooknum: nftables.ChainHookOutput, Priority: nftables.ChainPriorityFilter,
	})
	return c.Flush()
}

func chain() *nftables.Chain {
	return &nftables.Chain{
		Name: "output", Table: table, Type: nftables.ChainTypeFilter,
		Hooknum: nftables.ChainHookOutput, Priority: nftables.ChainPriorityFilter,
	}
}

// sync writes the ruleset: UDP to the test port goes to the queue, without the
// bypass flag so nothing is accepted behind our back.
func sync(mode string) error {
	c, err := nftables.New()
	if err != nil {
		return err
	}
	c.AddTable(table)
	if mode == "recreate" {
		c.DelTable(table)
		c.AddTable(table)
	}
	ch := c.AddChain(chain())
	if mode == "flushchain" {
		c.FlushChain(ch)
	}
	c.AddRule(&nftables.Rule{Table: table, Chain: ch, Exprs: []expr.Any{
		&expr.Meta{Key: expr.MetaKeyL4PROTO, Register: 1},
		&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []byte{syscall.IPPROTO_UDP}},
		&expr.Payload{DestRegister: 1, Base: expr.PayloadBaseTransportHeader, Offset: 2, Len: 2},
		&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []byte{byte(port >> 8), byte(port & 0xff)}},
		&expr.Queue{Num: queueNum},
	}})
	return c.Flush()
}

func main() {
	mode := flag.String("mode", "none", "none, recreate, flushchain or othertable")
	flag.Parse()

	if err := sync("recreate"); err != nil { // same starting state for every mode
		fmt.Fprintln(os.Stderr, "setup:", err)
		os.Exit(1)
	}
	defer func() {
		c, _ := nftables.New()
		c.DelTable(table)
		c.DelTable(other)
		_ = c.Flush()
	}()

	ctx, cancel := context.WithTimeout(context.Background(), duration+2*time.Second)
	defer cancel()

	nf, err := nfqueue.Open(&nfqueue.Config{
		NfQueue: queueNum, MaxPacketLen: 128, MaxQueueLen: 4096,
		Copymode: nfqueue.NfQnlCopyPacket, WriteTimeout: 100 * time.Millisecond,
	})
	if err != nil {
		fmt.Fprintln(os.Stderr, "nfqueue:", err)
		os.Exit(1)
	}
	defer nf.Close()

	var enoent atomic.Int64
	err = nf.RegisterWithErrorFunc(ctx,
		func(a nfqueue.Attribute) int {
			id := *a.PacketID
			go func() { // hold the verdict, then accept
				time.Sleep(hold)
				_ = nf.SetVerdict(id, nfqueue.NfAccept)
			}()
			return 0
		},
		func(e error) int {
			if errors.Is(e, syscall.ENOENT) {
				enoent.Add(1)
			}
			return 0
		})
	if err != nil {
		fmt.Fprintln(os.Stderr, "register:", err)
		os.Exit(1)
	}

	server, err := net.ListenUDP("udp4", &net.UDPAddr{IP: net.IPv4(127, 0, 0, 1), Port: port})
	if err != nil {
		fmt.Fprintln(os.Stderr, "listen:", err)
		os.Exit(1)
	}
	var received atomic.Int64
	go func() {
		buf := make([]byte, 64)
		for {
			if _, err := server.Read(buf); err != nil {
				return
			}
			received.Add(1)
		}
	}()

	stop := make(chan struct{})
	go func() {
		for {
			select {
			case <-stop:
				return
			case <-time.After(syncGap):
				var err error
				switch *mode {
				case "none":
				case "othertable":
					err = rebuildOther()
				default:
					err = sync(*mode)
				}
				if err != nil {
					fmt.Fprintln(os.Stderr, "sync:", err)
				}
			}
		}
	}()

	conn, err := net.DialUDP("udp4", nil, &net.UDPAddr{IP: net.IPv4(127, 0, 0, 1), Port: port})
	if err != nil {
		fmt.Fprintln(os.Stderr, "dial:", err)
		os.Exit(1)
	}
	var sent int64
	for end := time.Now().Add(duration); time.Now().Before(end); {
		if _, err := conn.Write([]byte("x")); err == nil {
			sent++
		}
		time.Sleep(sendGap)
	}
	close(stop)
	time.Sleep(hold + 500*time.Millisecond) // let the last verdicts land

	fmt.Printf("mode=%-10s sent=%d received=%d lost=%d enoent=%d
",
		*mode, sent, received.Load(), sent-received.Load(), enoent.Load())
}
```

```
# go.mod
module flushtest

go 1.26

require (
	github.com/florianl/go-nfqueue/v2 v2.1.0
	github.com/google/nftables v0.3.0
)

require (
	github.com/google/go-cmp v0.7.0 // indirect
	github.com/mdlayher/netlink v1.9.0 // indirect
	github.com/mdlayher/socket v0.5.1 // indirect
	golang.org/x/net v0.50.0 // indirect
	golang.org/x/sync v0.19.0 // indirect
	golang.org/x/sys v0.41.0 // indirect
)
```

```
docker run --rm -v "$PWD":/src -w /src golang:1.26 \
  sh -c 'go mod tidy && CGO_ENABLED=0 go build -o flushtest .'
for m in none othertable recreate flushchain; do
  docker run --rm --privileged -v "$PWD":/t alpine:3.22 /t/flushtest -mode $m
done
```

Output:

```
mode=none       sent=3599 received=3599 lost=0   enoent=0
mode=othertable sent=3643 received=3308 lost=335 enoent=335
mode=recreate   sent=3633 received=3295 lost=338 enoent=338
mode=flushchain sent=3634 received=3634 lost=0   enoent=0
```

`recreate` is what both sync functions do today. `flushchain` keeps the table and its base chain and replaces only the rule. `othertable` never touches the table the queue rule lives in: it rebuilds a second, unrelated table that has its own hooked chain and no queue rule at all, and the queue still loses the same share of packets, because the flush is namespace wide rather than per table. Every lost packet comes with exactly one ENOENT, and repeated runs of `recreate` stay within 335 to 365 lost packets.

**Environment**

- Linux 7.0.12, x86_64, in a privileged container. No Kubernetes involved.
- Go 1.26.8, `github.com/google/nftables v0.3.0`, `github.com/florianl/go-nfqueue/v2 v2.1.0`, matching this repo's `go.mod`.

**Related**

aojea/kindnet#158 and florianl/go-nfqueue#58 report the same error together with packet loss that was never explained. kindnet's `pkg/dnscache/agent.go` rebuilds its table the same way.

bendwyer:
#### What this PR does / why we need it

Deleting a table deletes its base chains, and deleting a base chain unregisters its netfilter hook, which makes the kernel drop every packet waiting in any nfqueue of the network namespace. Both `syncNFTablesRules` and the dnscache's `syncRules` rebuilt their table that way on every sync.

Keep the table and its base chains and replace only the rules and set elements, deleting chains and sets the current configuration no longer uses in the same transaction. The resulting ruleset is unchanged, which `TestNetworkPolicies_Resync` asserts by comparing a resync against a clean sync. `TestController_ResyncKeepsQueuedPackets` covers the packet loss: it receives 8 to 27 of 150 packets on main across four runs, and all 150 with this change.

`make test` and `make lint` pass. The reproducer, the kernel call path and the measurements are in the issue.

#### AI tool disclosure

This change was prepared with Claude Code. Every claim in the issue and here was checked against the v6.18 kernel sources cited there, and every number comes from a run of the reproducer or of the tests in this PR, not from the model's recollection. Happy to provide any further detail or to restructure the change if you would prefer a different approach.

Fixes #402


pkg/dataplane/controller.go
```diff
@@ -469,21 +470,26 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("can not start nftables:%v", err)
 	}
-	// add + delete + add for flushing all the table
+	// Keep the table and its base chains, and replace only their rules and set
+	// elements. Deleting a base chain unregisters its netfilter hook, and the
+	// kernel then drops every packet waiting in any nfqueue of the network namespace.
```
aojea, line 475: the operation is atomic, so the kernel should guarantee the there is not packet drops, IIUIC this is the canonical way of recreate a table from scratch without dataplane impact ... can you elaborate more on how did you hit this problem?

bendwyer: Thanks for taking a look. You are right that the swap is atomic and the ruleset never has a gap. What I ran into is narrower: packets already handed to userspace and waiting for a verdict sit in the queue instance rather than in the ruleset, and the `DelTable` step takes those out.

Even though the table is re-added in the same transaction, the delete still removes the old base chains, and removing a base chain unregisters its netfilter hook. That calls `nf_tables_unregister_hook`, which reaches `nf_queue_nf_hook_drop(net)` (`core.c:522`), which walks every queue instance in the namespace and calls `nfqnl_flush(inst, NULL, 0)` (`nfnetlink_queue.c:1184`). With `cmpfn == NULL` every entry matches, so each one is reinjected with `NF_DROP` (`:403`). (The line numbers are from v6.18.)

Because the flush is namespace wide, the queue that loses packets does not have to be the one being rebuilt. I added a mode to the reproducer in #402 that rebuilds a second, unrelated table with its own hooked chain and no queue rule, while traffic flows through the first table's queue:

```
mode=none       sent=3599 received=3599 lost=0   enoent=0
mode=othertable sent=3643 received=3308 lost=335 enoent=335
mode=recreate   sent=3633 received=3295 lost=338 enoent=338
mode=flushchain sent=3634 received=3634 lost=0   enoent=0
```

How we ran into it: on a four node cluster the dnscache's 30 second rebuild was dropping packets parked in the policy queue, around 9,900 a day, each one an ENOENT on a late verdict.

This is an AI assisted crack at fixing it, so if you would rather I approach it differently I am happy to follow your lead.

aojea:
/ok-to-test

This is awesome, I was chasing this problem for years, let's try to get this fixed

pkg/dataplane/controller.go
danwinship: This comment is assuming that the reader knows how the code _used to_ work. It should be rewritten to make sense to future readers, not to people reviewing this PR. It may make more sense if some or all of it comes later on in the function (eg, at the point where you actually delete those objects).

pkg/dataplane/controller.go
```diff
@@ -469,21 +470,26 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
 	if err != nil {
 		return err
 	}
 
+	staleChains, staleSets, err := staleTableObjects(table, divertAll, c.config.NetfilterBug1766Fix)
+	if err != nil {
+		return err
+	}
+
+	nft.AddTable(table)
```
danwinship: I think it makes more sense to keep the `AddTable()` right after the `table := ...`

pkg/dataplane/controller.go
```diff
@@ -923,6 +949,49 @@ func (c *Controller) addDNSRacersWorkaroundRules(nft *nftables.Conn, table *nfta
...
 }
 
+// staleTableObjects lists the chains and sets already in the table that a sync
+// with this configuration does not recreate, such as the prerouting chain after
+// NetfilterBug1766Fix is turned off. A table that does not exist yet has none.
+func staleTableObjects(table *nftables.Table, divertAll, netfilterBug1766Fix bool) ([]*nftables.Chain, []*nftables.Set, error) {
+	wantChains := map[string]bool{"postrouting": true, "input": true, "prerouting": netfilterBug1766Fix}
+	wantSets := map[string]bool{podV4IPsSet: !divertAll, podV6IPsSet: !divertAll}
+
+	nft, err := nftables.New()
```
danwinship: The caller already has one of these. You should pass that in rather than creating a new one. (It's more efficient, and in the future may reuse an existing netlink socket, etc.)

pkg/dataplane/controller.go
```diff
@@ -510,11 +516,19 @@ func (c *Controller) syncNFTablesRules(ctx context.Context) error {
...
 
-		if err := nft.AddSet(v4Set, elementsV4); err != nil {
-			return fmt.Errorf("failed to add Set %s : %v", v4Set.Name, err)
-		}
-		if err := nft.AddSet(v6Set, elementsV6); err != nil {
-			return fmt.Errorf("failed to add Set %s : %v", v6Set.Name, err)
+		for _, set := range []struct {
+			set      *nftables.Set
+			elements []nftables.SetElement
+		}{{v4Set, elementsV4}, {v6Set, elementsV6}} {
```
danwinship: not sure this is an improvement... if you want to avoid doing add/flush/addelements twice, you could add a helper for that

aojea:
I think we just need `FlushTable` https://wiki.nftables.org/wiki-nftables/index.php/Atomic_rule_replacement

https://github.com/google/nftables/blob/f9b52ed2ba653815684ff3e19a4d087ff0b87332/table.go#L151-L166

```go
// FlushTable removes all rules in all chains within the specified Table. See also
// https://wiki.nftables.org/wiki-nftables/index.php/Configuring_tables#Flushing_tables
func (cc *Conn) FlushTable(t *Table) {
```

aojea:
@bendwyer this is what my assistant gave me https://github.com/kubernetes-sigs/kube-network-policies/compare/main...aojea:kube-network-policies:nftables-atomic-rule-replacement?expand=1 , it looks pretty neat ... feel free to adopt it here, I think better to use Flush

aojea:
@bendwyer see https://github.com/kubernetes-sigs/kindnet/pull/149 for an example of a helper with tests that we can use to abstract the logic and make it reusable

danwinship:
> I think we just need `FlushTable`

No, that function is badly named. `nft flush table` just means "call `nft flush chain` on every chain in the table". It doesn't delete the chains, and it doesn't do anything at all with sets/maps/etc.

pkg/dataplane/controller.go
bendwyer: I rewrote this and the dnsagent one to describe what the code actually does. The explanation of the one-time recreate has been down to where it happens.

pkg/dataplane/controller.go
bendwyer: Done, it's back next to the table declaration.

pkg/dataplane/controller.go
bendwyer: This one solved itself now that we're using atomic rule replacement, so nothing opens a second connection any more.

pkg/dataplane/controller.go
bendwyer: I've pulled it into a small replaceSet helper and call it once per family instead.

bendwyer:
> @bendwyer see [kubernetes-sigs/kindnet#149](https://github.com/kubernetes-sigs/kindnet/pull/149) for an example of a helper with tests that we can use to abstract the logic and make it reusable

Thanks @aojea, I've taken your branch and credited you as co-author, and rebased it onto main. 

Two things: the set replacement is now a `replaceSet` helper, per @danwinship's comment, and a failed transaction clears `flushed` so the next sync recreates the table. Without that last bit, one good sync followed by a base chain whose hook or priority no longer matches, say from an old pod during a rolling upgrade, would fail every sync after it, since the kernel turns that update down with `-EEXIST`.

Your `TestNetworkPolicies_FirstSyncRecreatesTable` came along with the branch, so the first-sync behavior is covered too.

On the helper: `replaceSet` is the small local version of that. Happy to factor it the way you did in kindnet#149 if you want the same shape in both, though it might be easier once that one lands and the API has settled.

I amended rather than stacking commits, to keep the single commit this repo merges with. The force-push compare link should show what changed since your review.

bendwyer:
I ran an automated code review over this, and it turned up a gap the change leaves open. I checked it by hand afterwards, and I'm unsure how you would like it handled.

The one-time recreate still drops packets in every other nfqueue in the namespace, whenever a table from a previous process is still around. For the dnscache that is any restart, since nothing removes its table on shutdown.

In `pkg/dns/dnsagent.go` the queue opens at line 72 and the first `syncRules` runs at line 126, so the `DelTable` lands while queues are already live. The dataplane controller does it the other way round, initial sync at line 217 and `nfqueue.Open` at line 245, so its own queue is fine. The catch is that in the npa builds both run in the same namespace and `domainResolver.Run` starts first, so each one's first sync can drain the other's queue.

I measured it by holding packets in an independent queue and then starting `DomainCache.Run`, leaving a stale table behind to stand in for a previous process:

```
in flight before start: 74   lost: 73   enoent: 73
in flight before start: 77   lost: 76   enoent: 76
in flight before start: 76   lost: 75   enoent: 75
```

Without the stale table nothing is lost, since the delete then only applies to a table created in the same transaction. kindnet's dnscache looks to have the same ordering, queue at line 147 and the first sync later.

@aojea let me know how you would like to proceed. I can share the probe if it is useful.

aojea:
> The one-time recreate still drops packets in every other nfqueue in the namespace, whenever a table from a previous process is still around.

This looks something we should report to netfilter folks, I do not expect replacing a random table will impact packets on the queue from other table.

@danwinship @bendwyer what about casting the error and only do the Add/Delete/Add if the flush was not enough?

